PR #52, opened 2026-05-05
fix(api): correct agent workspace path and add duplicate guard

0xthoth:
## Summary
- Fix workspace path: removed extra `..` segment — was resolving to `/home/dev/agents/<id>/workspace` instead of `~/.claude-gateway/agents/<id>/workspace`
- Add file-level duplicate check inside `writeAgentsToConfig` — prevents race condition where two simultaneous POST requests write duplicate agent entries to config.json, which caused gateway config reload to fail

## Root Cause
`path.dirname(configPath)` = `~/.claude-gateway`, then joining `'..'` stepped up one level to `/home/dev` instead of staying in the gateway directory.

## Test Plan
- [ ] Create agent via POST `/api/v1/agents` → workspace appears at `~/.claude-gateway/agents/<id>/workspace`
- [ ] POST same agent ID twice quickly → second request returns 409, no duplicate in config.json
- [ ] Gateway config reloads cleanly after agent creation

### DIFF
--- a/src/agent/runner.ts
+++ b/src/agent/runner.ts
@@ -1141,11 +1141,22 @@ export class AgentRunner extends EventEmitter {
 
     const systemNote = buildApiSystemNote(opts.allowTools ?? false);
 
+    // Detect skill commands (same as channel message path)
+    const skillInvocation = detectSkillCommand(message, this.skillRegistry);
+    if (skillInvocation) {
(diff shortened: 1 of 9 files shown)


0xthoth:
Opened in wrong repo. Correct PR: https://github.com/0xthoth/claude-gateway/pull/9

PR #67, opened 2026-05-13
feat: add firstMessage field to SessionSummary

0xthoth:
## Summary
- Add `firstMessage` field to `SessionSummary` type in `src/history/types.ts`
- Populate it via an `ASC` subquery on `role = 'user'` in `listSessions()` in `src/history/db.ts`
- Additive change — does not modify `lastMessage`, existing fields, or any other API behavior

## Motivation
Expose the first user message in session list responses so clients can use it as a conversation label without fetching full chat history.

## Test plan
- [ ] `GET /v1/sessions` — verify `firstMessage` is populated for sessions that have at least one user message
- [ ] Sessions with no user messages return `firstMessage: ""`
- [ ] `lastMessage` and all other fields are unaffected
- [ ] No regression in existing session listing behavior

### DIFF
--- a/src/history/db.ts
+++ b/src/history/db.ts
@@ -266,7 +266,10 @@ export class HistoryDB {
         MAX(ts)   AS last_activity,
         (SELECT content FROM messages m2
          WHERE m2.session_id = m.session_id
-         ORDER BY ts DESC LIMIT 1) AS last_message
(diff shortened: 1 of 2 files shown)


0xthoth:
Cancelled — feature reverted

PR #80, opened 2026-05-15
fix(session): restore conversation context for API sessions on model-switch respawn

0xthoth:
## Summary

When an API user switches the model mid-conversation, the gateway spawns a new Claude process for the new model. Previously, the new process had no awareness of prior messages — every model switch felt like a fresh session with zero context. This PR fixes that by injecting the conversation history into the first message after a model-switch respawn.

---

## Problem

API sessions intentionally skip the activation prompt at spawn time. Sending a prompt at spawn races with the first API message turn and causes `sendApiMessage` to resolve with the wrong result. This is correct behaviour — but it also meant there was no path to inject conversation history into a newly-spawned session.

When a model switch triggered a respawn, the new Claude process started completely blind.

---

## Root Cause

`buildInitialPrompt()` always produced a single combined prompt (history + `CHANNELS_ACTIVATION_PROMPT`). For Telegram/Discord sessions this prompt is sent to `stdin` immediately at spawn. For API sessions nothing was sent — and there was no separate mechanism to carry the history-only portion forward.

---

## Solution

Split `buildInitialPrompt()` to return two values:

| Field | Content | Used by |
|-------|---------|---------|
| `prompt` | History + activation prompt | Telegram / Discord (unchanged) |
| `historyPrompt` | History only (no activation) | API model-switch respawn |

For API sessions that spawn with existing history, `historyPrompt` is stored as `pendingInitialPrompt` on the `SessionProcess`. On the **first** `sendMessage()` call after spawn, the history is prepended to the user's message text so Claude receives prior context within the same turn. `pendingInitialPrompt` is cleared immediately after — all subsequent messages are sent as plain text.

### Execution flow

```
User switches model (haiku → opus)
  ↓ gateway detects modelOverride changed
  ↓ existing session stopped, new one spawned with opus
  ↓ buildInitialPrompt() → historyPrompt = "[Conversation history with this user:\n...]"
  ↓ historyPrompt stored as pendingInitialPrompt (NOT sent to stdin at spawn)

User sends first message after switch
  ↓ sendMessage() detects pendingInitialPrompt
  ↓ fullText = pendingInitialPrompt + "\n\n" + userMessage
  ↓ combined turn written to Claude stdin
  ↓ pendingInitialPrompt cleared

User sends second message
  ↓ sendMessage() — pendingInitialPrompt is undefined → plain text sent
```

---

### Pros
- **Context continuity** — model switch no longer resets the conversation from Claude's perspective
- **Zero Telegram/Discord impact** — the code path for channel sessions is unchanged; only `historyPrompt` is a new return value alongside the existing `prompt`
- **No extra network round-trip** — history is injected inline with the user's first post-switch message, not as a separate preliminary turn
- **Respects existing history limits** — uses the same `MAX_HISTORY_MESSAGES` window and compaction-summary rescue logic already applied to channel sessions
- **Self-cleaning** — `pendingInitialPrompt` is always cleared on first use; no state leaks across messages

### Cons
- **First post-switch message is larger** — it carries both the history context and the user's actual message in a single turn. For long histories this increases token usage on that one turn
- **History is injected as plain text** — uses `[Conversation history with this user:\n...]` format (same as channel sessions), not structured API messages. The model sees context as a prompt prefix rather than native conversation turns
- **No history on very first API session spawn** — if the session has no prior messages (e.g. brand-new conversation), `historyPrompt` is null and nothing is injected (correct behaviour, but worth noting the asymmetry)

---

### `src/session/process.ts`
- Added `pendingInitialPrompt?: string` private field
- `buildInitialPrompt()` return type extended with `historyPrompt: string | null`
- API sessions store `historyPrompt` as `pendingInitialPrompt` instead of writing to stdin at spawn
- `sendMessage()` prepends `pendingInitialPrompt` before the first user turn and clears it

### `tests/unit/session-process.test.ts`
- Added new test suite `SessionProcess — API model-switch history injection`
- Updated U-SP-07, U-SP-SYS-01, U-SP-SYS-02 expectations to match current history format

---

## Tests

| ID | Description |
|----|-------------|
| U-SP-API-01 | API session does not write to stdin at spawn |
| U-SP-API-02 | History prepended to first `sendMessage()` after model-switch respawn |
| U-SP-API-03 | `pendingInitialPrompt` cleared after first message; subsequent messages are plain |
| U-SP-API-04 | API session with no prior history sends message as-is |

All existing tests pass.


0xthoth:
Closing — will reopen from a feature branch instead of fork:main

PR #85, opened 2026-05-18
fix: deleted API sessions reappear after gateway refetch

0xthoth:
## Problem

When deleting an API session, the session was only removed from the file index (JSON), while messages remained in the SQLite history database. Since `listSessions()` reads from SQLite, deleted sessions kept reappearing on every refetch.

Additionally, sessions created before the file index was introduced (legacy sessions) would throw "not found in index" errors on rename or delete operations.

## Changes

- `deleteApiSession`: now calls `historyDb.clearSession()` to purge messages from both storage layers (file index + SQLite)
- `deleteApiSession`: swallows "not found in index" errors for legacy sessions (session is effectively already gone)
- `updateApiSession`: calls `ensureApiSession` first to auto-register legacy sessions before attempting rename

## Test Plan

- [ ] Delete an API session via `DELETE /api/v1/agents/:agentId/sessions/:sessionId` — returns 204
- [ ] Confirm session does not reappear after `GET /api/v1/agents/:agentId/sessions` refetch
- [ ] Confirm legacy sessions (no file index entry) can be renamed and deleted without error


0xthoth:
Closing — branch contains unrelated commits from fork. Will resubmit with only the relevant fix commits.

PR #130, opened 2026-06-08
[Fix] Preserve SSE Stream On Client Disconnect

0xthoth:
## Problem

When an API client closes the SSE connection before the LLM finishes responding, the assistant reply is silently discarded — never saved to the database. On the next page reload, the conversation shows only the user message with no response, leaving the session in a permanently broken state.

**Root Cause:** `sendApiMessageStream` and `sendGreetingStream` returned the same `cleanup()` function for both "done" and "client disconnect" paths. Calling cleanup on disconnect marked the stream as `settled = true` and removed all listeners, so the LLM result that arrived later was dropped.

## Solution

Separate the two responsibilities:

- **`cleanup()`** — called internally when the LLM finishes (or errors). Saves the result to DB and frees the session slot.
- **`onClientDisconnect()`** — called by the SSE route handler on `res.on('close')`. Sets `clientGone = true` to silence all outbound SSE writes, but does **not** stop the stream or remove the `output` listener. The LLM continues processing server-side and cleanup runs normally when it finishes.

This ensures the assistant reply is always persisted regardless of whether the client is still connected.

## Changes Made

- `src/agent/runner.ts`: introduce `clientGone` flag in `sendApiMessageStream` and `sendGreetingStream`; guard all `callbacks.onChunk()` calls with `if (!clientGone)`; return `onClientDisconnect` instead of `cleanup`
- `src/history/db.ts`: add `last_message_role` subquery to `listSessions()` SQL — returns the role of the most recent message per session
- `src/history/types.ts`: add `lastMessageRole` field to `SessionSummary`
- `src/api/router.ts`: expose `lastMessageRole` in `GET /v1/agents/:agentId/sessions` response
- `tests/integration/chat-history-api.test.ts`: add I-HIST-13 — verifies assistant reply is saved after SSE client disconnects mid-stream
- `tests/unit/history-cleanup.test.ts`: add two cases for `lastMessageRole` — last message from user and from assistant

## How to Test

1. Start gateway and send a message via `POST /v1/agents/:id/messages?stream=true`
2. Close the SSE connection immediately after receiving the first chunk
3. Wait a few seconds for the LLM to finish
4. Fetch `GET /v1/agents/:id/chats/:chatId/messages`

**Expected:** the assistant reply is present in message history

For `lastMessageRole`:
1. Call `GET /v1/agents/:id/sessions`
2. Check the `lastMessageRole` field on each session

**Expected:** `"user"` if the last message is from the user (awaiting response), `"assistant"` if the last message is a reply

## Checklist
- [x] Reviewed my own code
- [x] Tests pass (1,678/1,678)
- [x] No console errors


0xthoth:
Superseded by #131 — clean cherry-pick onto upstream/main with only the relevant 6 commits.

PR #174, opened 2026-06-24
fix(session): defer history+activation to first user turn to prevent double-response on PTY-shell restart

0xMaxMa:
## Problem

When a PTY-shell session (headless=false) dies while an interactive menu is pending (e.g. ExitPlanMode Pre-flight Summary) and the subprocess auto-restarts:

1. `buildInitialPrompt()` loads the conversation history (which now includes the user's latest message, e.g. "Y", because `appendTelegramMessage` runs before `getOrSpawnSession`)
2. The old code sent **history + activation prompt immediately** to stdin → this becomes **Turn 1** → Claude responds (Response 1)
3. Then `sendMessage(channelXml)` delivers the user's actual reply as **Turn 2** → Claude responds again (Response 2)

Result: two separate responses forwarded to the Telegram channel.

## Fix

For non-API sessions **with history**, store the combined `history + CHANNELS_ACTIVATION_PROMPT` in `pendingInitialPrompt` instead of writing to stdin immediately.

`sendMessage()` already checks `pendingInitialPrompt` and prepends it to the first user message, bundling everything into a **single stdin turn**:
```
[history + activation]\n\n[user's channelXml]
```

Claude processes this as one turn → one response. Fresh sessions (no history) continue to send the activation prompt immediately, so there is no regression on first-ever messages.

## Files changed

| File | Change |
|------|--------|
| `src/session/process.ts` | Defer history+activation to `pendingInitialPrompt` for non-API sessions with history |
| `tests/unit/session-process.test.ts` | 10 tests updated to check `pendingInitialPrompt` instead of `stdin.write` |
| `tests/helpers/mock-claude-mcp.js` | Fix regex (`^` removed) and check first line for embedded channel XML |

## Testing

- All 80 tests in `session-process.test.ts` pass ✅
- `pipeline-mcp.test.ts` Step 3b passes ✅  
- `phase-manual.test.ts` failures (P3-10, P3-11, P3-12, P4-12) are **pre-existing** — confirmed by running on `main` before this change

### DIFF
--- a/src/session/process.ts
+++ b/src/session/process.ts
@@ -493,11 +493,22 @@ export class SessionProcess extends EventEmitter {
     // first API turn and cause sendApiMessage to resolve with the wrong result.
     // Instead, if there is conversation history to restore (model-switch respawn),
     // stash it in pendingInitialPrompt so sendMessage() prepends it to the first turn.
+    //
+    // Non-API sessions with history also use pendingInitialPrompt to avoid a
+    // double-response bug: if the session died while an interactive menu was pending
+    // (e.g. ExitPlanMode Pre-flight Summary), sending history + activation immediately
+    // makes Claude respond to that context as Turn 1, then the user's reply (e.g. "Y")
+    // becomes Turn 2 — two separate responses forwarded to the channel. Deferring
+    // history to sendMessage() bundles [history + activation + user reply] into a
+    // single turn so Claude produces exactly one response.
     if (this.source !== 'api') {
-      const initialPrompt = historyPrompt
-        ? `${historyPrompt}\n\n${CHANNELS_ACTIVATION_PROMPT}`
-        : CHANNELS_ACTIVATION_PROMPT;
-      proc.stdin?.write(SessionProcess.toStreamJsonTurn(initialPrompt) + '\n');
+      if (historyPrompt) {
+        // Has history: defer to first incoming user message to prevent double-response.
+        this.pendingInitialPrompt = `${historyPrompt}\n\n${CHANNELS_ACTIVATION_PROMPT}`;
+      } else {
+        // No history: send activation-only prompt immediately (fresh session).
+        proc.stdin?.write(SessionProcess.toStreamJsonTurn(CHANNELS_ACTIVATION_PROMPT) + '\n');
+      }
     } else if (historyPrompt) {
       this.pendingInitialPrompt = historyPrompt;
     }
--- a/tests/helpers/mock-claude-mcp.js
+++ b/tests/helpers/mock-claude-mcp.js
@@ -125,7 +125,10 @@ function parseChannelFromStdin(line) {
   try {
     const obj = JSON.parse(line)
     const text = obj?.message?.content?.[0]?.text ?? ''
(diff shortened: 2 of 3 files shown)


0xMaxMa:
🔍 Code Review — #174 fix(session): defer history+activation to first user turn (self-review)

📊 Score: 8.5 / 10.0
- Code Quality:          8.5
- Correctness:           8.5
- Production Readiness:  8.5

---

## 💪 Strengths

- **Reuses existing infrastructure**: `pendingInitialPrompt` was already wired for API sessions — the fix elegantly extends that pattern to non-API sessions without introducing new fields or mechanisms
- **Comment captures the subtle invariant**: the block explaining *why* `appendTelegramMessage` runs before `getOrSpawnSession` (and why that makes `historyPrompt` non-null for the first-ever message) is exactly the kind of non-obvious timing contract that rots without documentation
- **Dual-layer fix**: both the production code path (`process.ts`) and the test helper (`mock-claude-mcp.js`) are updated coherently — no loose ends where one side assumed the old behavior
- **Pre-existing failures confirmed rigorously**: `git stash` was used to prove P3-10/11/12/P4-12 were pre-existing, keeping the PR diff clean

---

## 🔴 Critical (0)
*(none)*

---

## 🟡 Medium (0)
*(none)*

---

## 🟢 Minor (3)

**1. `tests/unit/session-process.test.ts` — Private field access via double type-cast**

All 10 updated tests access `pendingInitialPrompt` through `(sp as unknown as { pendingInitialPrompt?: string })`. If `pendingInitialPrompt` is renamed in `process.ts`, TypeScript won't catch the stale reference in the tests — they'll silently return `''` and the assertions will pass vacuously.

→ Not urgent, but consider adding a `getPendingInitialPrompt(): string | undefined` test accessor or making the field `protected` so subclasses (and test utilities) can reach it type-safely.

**2. `tests/helpers/mock-claude-mcp.js:126` — Anchored regex removed, could match stale channel XML in history**

Removing `^` lets `/<channel([^>]*)>([\s\S]*?)<\/channel>/` match the *first* `<channel>` anywhere in the bundled text. In tests with simple synthetic history (`"User: Hello"`, `"Assistant: Hi"`) this is harmless. But if history ever contains a literal channel-XML string (e.g., from a logged prior conversation), `parseChannelFromStdin` would extract the wrong message — the one from history rather than the current user turn.

→ For robustness, match specifically the last `<channel>` occurrence (the current turn is always appended last), or anchor to the end of the text: use `[\s\S]*(<channel([^>]*)>([\s\S]*?)<\/channel>)` with a greedy outer group.

**3. No integration test for the actual double-response scenario**

The fix is motivated by a specific failure mode: session dies with pending menu → user sends "Y" → two responses forwarded. The unit tests verify the *mechanism* (`pendingInitialPrompt` is set, stdin.write is not called) but not the end-to-end behavior (exactly one reply reaches the channel). The pipeline-mcp Step 3b test covers a related path (bundled first turn processed correctly) but not the restart-with-pending-menu case.

→ A targeted integration test that simulates a subprocess restart with prior history and verifies exactly one `routeResponse` call would lock in the invariant. Not blocking, but worth a follow-up issue.

---

## 🏁 Verdict: ✅ Ready to merge

No critical or medium issues. Three minor points — none block shipment.

PR #175, opened 2026-06-25
fix(pty-shell): prevent stuck PTY input after /stop (selectMenuOption + typeAndSubmit)

0xMaxMa:
## Root Cause

Two gaps remained after #167 that allowed text to linger in the PTY input prompt after `/stop`, causing the next message to be appended to stale text and submitted together.

### Bug 1 — `selectMenuOption` has no interrupt check (primary culprit)

When the user selects a menu option (ExitPlanMode / AskUserQuestion) and then issues `/stop` during `MENU_SELECT_ENTER_DELAY_MS`:

1. Digit `"1"` or `"2"` is written to PTY input
2. SIGINT → ESC dismisses the menu (no Enter sent, so the digit isn't submitted)
3. **No interrupt check existed** → digit is never erased
4. Digit lingers in the input prompt
5. Next message pasted by `typeAndSubmit` produces `"1<new message>"` → Enter → Claude receives both

This is why the fix from #167 didn't catch this: #167 only fixed `typeAndSubmit`, not `selectMenuOption`.

### Bug 2 — `typeAndSubmit` pre-paste guard missing

If SIGINT fires **during `writeChunked`'s inter-chunk delays** (10 ms each, only for messages > 8 KB), ESC is injected into the middle of the bracketed paste stream. The existing post-paste Ctrl+U has to clear a partially-corrupt input. Adding a pre-paste guard (`this.interrupting || !this.turn`) eliminates the write entirely in the narrow window where SIGINT fires before the paste begins.

### Bug 3 — Ctrl+U only, no fallback

The Ink-based Claude Code TUI may not handle `\x15` (Ctrl+U) in all configurations. Upgrading to `\x01\x0b\x15` (Ctrl+A → go to start, Ctrl+K → kill to end, Ctrl+U → kill-line fallback) ensures clearing works regardless of cursor position and TUI readline variant.

## Changes

- **`selectMenuOption`**: interrupt check after delay → clear digit with `\x01\x0b\x15`; return without setting `submittedAt`
- **`typeAndSubmit`**: pre-paste guard at top; post-paste clearing upgraded to `\x01\x0b\x15`

## Test plan

- [ ] tsc clean — ✅ (`npm run build` passes)
- [ ] All 83 pty-shell tests pass — ✅
- [ ] Manual: Send message to agent → `/stop` mid-response → next message arrives clean (no prepended stale text)
- [ ] Manual: ExitPlanMode shows menu → press `1` → `/stop` before confirm → next message arrives clean

> Note: `Driver.typeAndSubmit` and `Driver.selectMenuOption` are private methods; the existing test suite has no PTY mock to test them directly. Behavioral coverage is through the integration (pipeline-mcp) tests.

### DIFF
--- a/src/shell/claude-pty-shell.ts
+++ b/src/shell/claude-pty-shell.ts
@@ -387,6 +387,12 @@ class Driver {
   }
 
   private async typeAndSubmit(text: string): Promise<void> {
+    // Pre-paste guard: if SIGINT already fired between beginTurn() and here, skip the
+    // paste entirely — no text lands in the PTY input so no clearing is needed.
+    if (this.interrupting || !this.turn) {
+      this.queue.length = 0;
+      return;
+    }
     if (NO_BRACKETED_PASTE) {
       // Fallback: sanitizeUserText() strips all CR, so '\r' below is the only
       // submit trigger — safe for multiline text without bracketed paste.
@@ -403,7 +409,10 @@ class Driver {
     // Also clear the queue so any messages queued behind this one don't surface
     // after the turn is abandoned (matches the SIGINT queue-drop logic above).
     if (this.interrupting || !this.turn) {
-      this.host.writeRaw('\x15'); // Ctrl+U: clear input line
+      // Belt-and-suspenders: Ctrl+A (home) + Ctrl+K (kill-to-end) works regardless
+      // of cursor position; Ctrl+U is the readline fallback. Together they cover
+      // Ink TUI implementations that may only handle a subset of these sequences.
+      this.host.writeRaw('\x01\x0b\x15'); // Ctrl+A + Ctrl+K + Ctrl+U
       this.queue.length = 0;
       return;
     }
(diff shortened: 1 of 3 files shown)


0xMaxMa:
🔍 Code Review — #175 fix(pty-shell): prevent stuck PTY input after /stop (self-review)

📊 Score: 8.5 / 10.0
- Code Quality:         8.5
- Correctness:          8.0
- Production Readiness: 9.0

💪 Strengths
- `selectMenuOption` interrupt check fills the exact gap PR #167 missed — digit `1`/`2` written before the 250ms delay now gets erased by Ctrl+U if SIGINT arrives during that window
- Pre-paste guard in `typeAndSubmit` adds clean defense-in-depth; early return before any PTY write means zero text to clear
- `mock-claude-tui.js` is a solid, self-contained test double — 4-state bracketed-paste parser, clear-screen idle transition, and JSONL transcript writing all mirror real Ink TUI behavior precisely
- Test structure (baseline → mid-paste interrupt → post-settle) is logical and the assertion style (log-file polling) is reliable

🔴 Critical (0)
(none)

🟡 Medium (0)
(none)

🟢 Minor (3)
1. `tests/integration/pty-stop-stuck-input.test.ts` — No E2E test for `selectMenuOption` interrupt path, even though the PR describes it as "the primary culprit (Bug 1)". The three tests only exercise `typeAndSubmit`. A test that sends a menu option digit then fires SIGINT within 250ms would close the coverage gap.
2. `tests/integration/pty-stop-stuck-input.test.ts:131` — `process.kill(wrapper.pid!, 'SIGINT')` uses a non-null assertion. If the wrapper fails to spawn, `pid` is `undefined` and this throws outside Jest's catch, leaving the test suite in an undefined state. Guard with `if (wrapper.pid) process.kill(wrapper.pid, 'SIGINT')`.
3. PR description "Changes" section still mentions `\x01\x0b\x15` (Ctrl+A → Ctrl+K → Ctrl+U) but the actual code uses `\x15` (Bug 3 was reverted). Stale description will confuse future readers tracing the commit history.

🏁 Verdict: ✅ Ready to merge

PR #181, opened 2026-07-01
fix(pty): replace transcript-gated menu detection with a behavioral keypress probe

0xMaxMa:
## Summary
- Replaces screen-regex menu/permission detection (`detectMenu()`, `detectPermissionPrompt()`) and the transcript `turn.menuToolSeen` gate with a direct behavioral probe: send Down (fallback Up if already at the last option), diff the screen, confirm liveness before bridging `AskUserQuestion`/`ExitPlanMode` prompts to chat.
- Fixes a silent-hang class of bug: when the transcript tailer didn't confirm the `tool_use` in time (or it was nested under a sub-agent block the tailer doesn't attribute to the main turn), the old gate fail-safed to "don't bridge anything," leaving a live interactive prompt on the TUI invisible to the user.
- Restores input state after an Up-fallback probe that accidentally triggers shell history recall (auto-sends a compensating Down).
- Removes `detectRequestTooLarge()` (dead code, superseded by the transcript `<synthetic>` signal per its own doc comment) and the tests that only existed to cover it.

## Post-review hardening (second commit)
Adversarial review of the first commit surfaced 7 findings; all are fixed:
- **F1 (fake-menu bridge):** a raw screen-text diff counted *any* change as a menu reaction, so Up-fallback history recall plus static menu-shaped scrollback (e.g. a markdown `> 1.` blockquote) could bridge a fabricated menu. The probe now parses the prompt from both the before and after snapshots and bridges only when the same-shaped prompt's ❯-highlighted row **moved** (`confirmProbeReaction()`); the caret regex accepts only the real U+276F caret, never ASCII `>`. The restorative Down also fires for parseable-but-unmoved changes.
- **F2:** `tick()` returns before `maybeHandleDialog()` while a probe round is in flight — the dialog auto-accept `'2'` can no longer interleave with probe arrows.
- **F3:** `runProbe()` aborts at every resume point when a `/stop` interrupt is settling; no round starts mid-interrupt.
- **F4:** `interactivePromptBlocking()` scans the full viewport (same scope as the reader), so a menu high on a sparse screen can't lose its confirming Enter after a selection.
- **F5:** the busy-resume branch no longer resets the probe round budget while a round is in flight.
- **F6:** removed AI signatures from the commit message and this description per repo `CLAUDE.md`.
- **F7:** the blocking check reuses `CARET_OPTION_RE` instead of an inline duplicate.

## Test plan
- [x] TypeCheck clean
- [x] Full unit test suite (`pty-shell.test.ts`) — 96/96 passing (new: `confirmProbeReaction()`, highlighted-index extraction, blockquote-caret rejection, full-viewport blocking scan)
- [x] Integration suite (`tests/integration/pty-menu-probe.test.ts`) — 7 scenarios passing; new `RECALL_FAKEMENU` scenario reproduces the F1 fake-menu sequence and fails against the first commit
- [x] Full repo test suite — 2104/2113 passing; 4 failures are pre-existing on `main` (`phase-manual.test.ts` Phase 3/4 `/status` 500s) and 1 (`api-router` T8b) is load-flaky, passing in isolation — both unrelated to this change
- [ ] Manual verification against a live TUI via VNC (planning-61 Task 5 — not yet done, flagged to the requester)

## Post-review hardening round 2 (third commit)
A second adversarial review of the hardened commit surfaced 7 more findings; all are fixed:
- **R2-1 (session-wedge regression):** the F4 full-viewport scan let a user draft starting with `N.` (rendered as `❯ 1. …` on the input line) suppress the swallowed-Enter retry, wedging the turn until the watchdog killed the session. `interactivePromptBlocking()` now scopes caret matching away from the input line.
- **R2-2 (forged highlight):** `caretOptionIndex()` read the highlight from anywhere on screen, so digit-leading history text recalled into the input line could still forge a highlight move. The highlight is now read only from the option-run region.
- **R2-3 (orphaned recall):** probe abandon paths (busy-race, turn-end, interrupt mid-settle) skipped the restorative Down after an Up fallback, leaving recalled history text prepended to the next prompt. The probe is now a tick-driven state machine (`advanceProbe`) that guarantees restoration on every exit path.
- **R2-4:** probing is gated on `sawBusy`/records so a swallowed-Enter stall is retried instead of probed.
- **R2-5:** multi-line history recalls are fully restored (one Down per recalled line).
- **R2-6/7 (cleanup):** the detached probe coroutine that caused the F2/F3/F5 race class is gone (subsumed by the state machine), and ~150 lines of duplicated test scaffolding are extracted into `tests/helpers/mock-tui-core.js` + `tests/helpers/pty-harness.ts`.

New regression coverage: digit-leading recall cannot forge a highlight move; turn-end mid-probe still clears the recall; a numbered draft with a swallowed Enter is retried, not probed. Unit 99/99, integration 13/13.

### DIFF
--- a/src/session/process.ts
+++ b/src/session/process.ts
@@ -227,9 +227,8 @@ export class SessionProcess extends EventEmitter {
       // Defense-in-depth: defang any verbatim 32MB-overlay text captured into past
       // messages before it is re-typed into the TUI. The primary fix routes the real
       // error off the screen-scraper to the transcript's `<synthetic>` record (see
-      // TranscriptTailer.onRequestTooLarge), so detectRequestTooLarge() is no longer
-      // wired into the trigger — but neutralizing the re-injected copy keeps the
-      // poisoned overlay text off the screen entirely and out of any future scraper.
+      // TranscriptTailer.onRequestTooLarge) — but neutralizing the re-injected copy
+      // keeps the poisoned overlay text off the screen entirely and out of any future scraper.
       historyPrompt: `[Conversation history with this user:\n${neutralizeTuiTriggers(historyText)}]`,
       loadedAtSpawn,
       archivedCount,
--- a/src/shell/claude-pty-shell.ts
+++ b/src/shell/claude-pty-shell.ts
@@ -14,13 +14,13 @@ import * as fs from 'fs';
 import * as net from 'net';
 import * as readline from 'readline';
 import { translateArgs, sanitizeUserText } from './args';
-import { ScreenModel, MenuOption, parseMenuChoice, formatMenuPrompt, formatPermissionPrompt, extractChannelContent, isPtyActivelyWorking } from './screen';
+import { ScreenModel, MenuOption, parseMenuChoice, formatMenuPrompt, formatPermissionPrompt, extractChannelContent, isPtyActivelyWorking, parseInteractivePrompt } from './screen';
 import { PtyHost } from './pty-host';
 import { TranscriptTailer, AssistantRecord, UsageInfo } from './tailer';
 import { ProtocolEmitter } from './emitter';
 import { preTrustWorkspace, checkAuthStatus } from './trust';
 import { decideMenuCancel } from './menu-cancel';
-import { decideProbeAttempt, ProbeState, PROBE_KEY_DOWN, PROBE_KEY_UP, PROBE_SETTLE_MS } from './menu-probe';
+import { decideProbeAttempt, confirmProbeReaction, ProbeState, PROBE_KEY_DOWN, PROBE_KEY_UP, PROBE_SETTLE_MS } from './menu-probe';
 
 const POLL_MS = 200;
 const STARTUP_QUIET_MS = 600;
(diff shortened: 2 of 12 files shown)


0xMaxMa:
## Task 5 — Live TUI verification results

Gateway restarted from this branch's build (verified `menu-probe.js` in running `dist/`). Tested via real gateway API sessions (agent `shadow`, PTY mode) and a directly-spawned wrapper harness (`PTY_SHELL_DEBUG=1`, fresh scratch workspace).

| # | Scenario | Result |
|---|----------|--------|
| 1 | AskUserQuestion (4 options) | ✅ PASS — `interactive menu confirmed (6 options, highlight 1→2)`, bridged in one probe round; digit reply "2" selected the right option even with the probe-moved highlight |
| 2 | ExitPlanMode plan approval | ✅ PASS — 3-option approval menu confirmed (highlight 1→2) and bridged; "1" approved and the plan executed. Resolves Open Question 2 (same list shape, one reader suffices) |
| 3 | Last-option highlight → Up fallback | ⚪ Not forceable live (fresh menus always start at row 1 with room below; no external raw-key access). Integration mock-TUI scenarios remain the authority |
| 3a | Up-recall input restoration | ⚪ Not forceable live (same reason). Covered by RECALL_NONMENU / RECALL_FAKEMENU / turn-end-mid-probe integration scenarios |
| 4 | Normal turn, no menu | ✅ PASS — 51s multi-step + 8s multi-tool turns: zero probe/menu events, clean output, no stray keystrokes |
| 5 | Dangerous-rm permission prompt | ⚪ Not reproducible on current Claude Code — live `rm -rf` of /tmp and nonexistent ~/ paths ran with no confirmation under built-in `--dangerously-skip-permissions` (catastrophic patterns not live-tested, obviously). Reader stays fixture-covered; passive behavior identical to main |
| 6 | Bypass-permissions startup dialog | ✅ PASS by absence + code path — dialog no longer renders (workspace pre-trusted, bypass accepted account-wide); all live sessions reached `session_idle` cleanly, and the probe cannot fire pre-ready (requires an active turn). Auto-accept path untouched by this PR |

Timing (Open Question 3): every live menu confirmed and bridged in **one probe round** — `PROBE_SETTLE_MS=200` / `PROBE_MAX_ROUNDS=3` need no tuning.

Full details appended to `planning-61-interactive-prompt-probe.md` (Task 5 — Live Verification Results).

PR #213, opened 2026-07-20
feat(history): ascending (seek-forward) message pagination for jump-to-date

### LINKED ISSUE #211: feat(history): ascending (seek-forward) message pagination for jump-to-date
## Context

A chat frontend gained a **jump-to-date** feature: clicking a day separator opens a calendar; picking a date scrolls the thread to that day's first message.

Today the frontend has no way to *seek* to a date, so it walks older history **one page at a time** — repeatedly calling `GET …/messages?limit=20&before=<cursor>` until the target day is paged in. For a date a few months back on a busy chat that's dozens of sequential round-trips + a large DOM before the jump lands. It's `O(history)`, not `O(1)`.

## Current API

`GET /api/v1/agents/:agentId/chats/:chatId/messages?limit&before&after&session_id`
→ `HistoryDb.getMessages()` (`src/history/db.ts:169`)

`PaginationOpts` (`src/history/types.ts`) already supports both bounds:
- `before` → `ts < ?`
- `after`  → `ts > ?`

**But the query is always `ORDER BY ts DESC` (`db.ts:194`) with `LIMIT limit+1`.** So `after=<t>` returns the **newest** `limit` messages after `t`, descending — *not* the messages **at/just after** day X in ascending order. There's no way to fetch "the page that starts at day X and reads forward" in a single call, which is exactly what a date-seek needs.

## Ask

Add an optional **sort direction** to the messages endpoint so a client can seek-forward from a timestamp:

- New `PaginationOpts.order?: 'asc' | 'desc'` (default `'desc'`, fully backward-compatible).
- When `order='asc'`, `getMessages` orders `ts ASC`; combined with `after=<startOfDay(X) − 1ms>&limit=N` it returns day X's **first N messages ascending** in one round-trip → the frontend anchors the jump immediately.
- `nextCursor` semantics for `asc`: the last (newest) `ts` returned, so the client can continue forward if needed.

## Scope

- `src/history/db.ts` — conditional `ORDER BY ts ASC|DESC` in `getMessages`.
- `src/history/types.ts` — add `order` to `PaginationOpts`.
- `src/api/router.ts:~1497` — parse `order` query param (`GET …/messages`).
- `API.md` — document `order`.
- Unit test in `tests/unit/api` for asc ordering + `after` bound + `hasMore`/`nextCursor`.

## Out of scope

Frontend consumption (rewriting the jump machine to call seek once instead of the page-by-page loop) — tracked as a separate frontend issue.

## Why now

Unblocks a true one-round-trip jump-to-date and removes the current sequential-paging cost. Pairs with the sibling "active-days" endpoint that powers the calendar's has-history indicators.

yuuyui:
## Summary

Adds ascending (seek-forward) message pagination to the history API so a **jump-to-date** action can seek forward from a timestamp in `O(1)` per page, instead of walking backward from newest through the whole history.

Fully backward-compatible: the default ordering is unchanged (`desc`); `asc` is opt-in via a new query param.

Closes #211

## Changes (6 files, +243 / −4)

| File | Change |
|------|--------|
| `src/history/types.ts` | Add `order?: 'asc' \| 'desc'` to `PaginationOpts` |
| `src/history/db.ts` | `getMessages` applies a **whitelisted** `ORDER BY ts ${order}` (default `desc`), tiebroken by `id` so rows sharing a `ts` get a deterministic, chronology-consistent order; `nextCursor` / `hasMore` / `limit+1` logic reused unchanged |
| `src/api/router.ts` | Parse `?order=asc` on the messages endpoint; validated against the whitelist before it reaches SQL, and a non-string (repeated/structured) param is rejected with `400` rather than throwing |
| `API.md` | Document the `order` param with a seek-forward example |
| `tests/unit/history-db.test.ts` | 13 new unit tests covering asc/desc ordering, cursor paging in both directions, same-`ts` id tiebreak, and the default |
| `tests/integration/chat-history-api.test.ts` | Integration test I-HIST-14: order asc/desc, case-insensitivity, and `400` on an invalid or repeated `order` |

## Design notes

- **No SQL injection surface** — `order` is validated against `{'asc','desc'}` and never interpolated raw; at the DB layer anything else falls back to `desc`.
- **Invalid input at the HTTP layer returns `400`** — an explicit `order` that isn't `asc`/`desc` (including a repeated/array param that Express parses as a non-string) is rejected so client typos surface instead of silently defaulting.
- **Cursor semantics unchanged** — `nextCursor = slice[last].ts`; continue forward via `after` for `asc`, backward via `before` for `desc`, so no cursor special-casing was needed. Known limitation: because the cursor is `ts`-only, a page boundary landing between equal-`ts` rows can still skip the tied remainder — fully closing that needs a composite `(ts, id)` cursor and is tracked as a separate follow-up.

## Testing

- `tsc` clean
- `history-db` **28/28** (incl. 13 new) · `api-router` **98/98** · `chat-history` integration — all green
- Full suite: pre-existing failures are confined to unrelated subsystems (PTY, telegram-plugin, mcp-pipeline, skills-runtime) — environmental integration tests needing real subprocess/timing; none touch history / messages / pagination.

## Scope

Backend API only. The frontend jump-to-date consumer is a separate issue and out of scope here.

### DIFF
--- a/API.md
+++ b/API.md
@@ -2367,7 +2367,7 @@ curl -H "X-Api-Key: my-secret-key-123" \
 
 ### GET /api/v1/agents/:agentId/chats/:chatId/messages
 
-Paginated message history (cursor-based). Returns messages in reverse chronological order.
+Paginated message history (cursor-based). Returns messages in reverse chronological order by default; pass `order=asc` to read forward.
 
 **Query parameters:**
 
@@ -2377,6 +2377,7 @@ Paginated message history (cursor-based). Returns messages in reverse chronologi
 | `before` | Return messages before this timestamp (ms) |
 | `after` | Return messages after this timestamp (ms) |
 | `session_id` | Filter to a specific session |
+| `order` | `asc` reads forward (oldest→newest) from `after`; default `desc` (newest→oldest) |
 
 ```bash
 curl -H "X-Api-Key: my-secret-key-123" \
@@ -2392,6 +2393,15 @@ curl -H "X-Api-Key: my-secret-key-123" \
   "hasMore": false
 }
 ```
+
+**Seek-forward example** (jump to a date and read that day's first messages in one round-trip):
+
+```bash
+curl -H "X-Api-Key: my-secret-key-123" \
+  "http://localhost:10850/api/v1/agents/alfred/chats/telegram-<CHAT_ID>/messages?order=asc&after=<startOfDay-1>&limit=20" | jq
+```
+
+`nextCursor` continues forward via `after` when `order=asc` (vs. `before` for the default `desc`).
 
 ---
 
--- a/API.md
+++ b/API.md
@@ -2377,7 +2377,7 @@ Paginated message history (cursor-based). Returns messages in reverse chronologi
 | `before` | Return messages before this timestamp (ms) |
 | `after` | Return messages after this timestamp (ms) |
 | `session_id` | Filter to a specific session |
-| `order` | `asc` reads forward (oldest→newest) from `after`; default `desc` (newest→oldest) |
+| `order` | `asc` reads forward (oldest→newest) from `after`; `desc` (default) reads newest→oldest. Case-insensitive; any other value returns `400`. |
 
 ```bash
 curl -H "X-Api-Key: my-secret-key-123" \
--- a/src/api/router.ts
+++ b/src/api/router.ts
@@ -2034,7 +2034,7 @@ export function createApiRouter(
   /**
    * GET /api/v1/agents/:agentId/chats/:chatId/messages
    * Paginated message history (cursor-based).
-   * Query: limit, before (ts ms), after (ts ms), session_id
+   * Query: limit, before (ts ms), after (ts ms), session_id, order (asc|desc, default desc)
    */
   router.get('/v1/agents/:agentId/chats/:chatId/messages', auth, (req: Request, res: Response) => {
     const { agentId, chatId } = req.params as { agentId: string; chatId: string };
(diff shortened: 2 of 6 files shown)


yuuyui:
🔍 **Code Review — #213 feat(history): ascending seek-forward pagination** (self-review)

📊 **Score: 9.3 / 10.0**
- Code Quality: 9.5
- Correctness: 9.0
- Production Readiness: 9.5

💪 **Strengths**
- **Two-layer whitelist** (`router.ts:2053` + `db.ts:187`) — raw `order` never reaches SQL string interpolation; only the literal `ASC`/`DESC` constant does. Textbook injection prevention.
- **Cursor logic correctly reused unchanged for both directions** — `nextCursor = last ts of slice` is right for `asc` (newest) and `desc` (oldest). Verified by a contiguity test asserting no gap / no overlap across pages.
- **Behavior-focused tests**, not mock-assertions: page contiguity, bounded windows (`before`+`after`), empty/unknown-chat, `sessionId` composition.
- **Minimal, backward-compatible surface** (+144/−4) — default `desc` path byte-for-byte unchanged; no over-engineering.

♻️ **Reuse & Duplication (0)**
- No duplication or over-engineering found. Change adds one field + one whitelisted ternary in two layers; no new helper/component/abstraction to reinvent.

🔴 **Critical (0)**
(none)

🟡 **Medium (0)**
(none)

🟢 **Minor (2)**
1. `src/history/db.ts:169` — Cursor is `ts`-only with strict `>`/`<`, so at an exact `ts` collision straddling a page boundary one message can be skipped. **Pre-existing and symmetric** — the `desc` path on `main` has the identical property; this PR faithfully mirrors it and does **not** introduce or worsen it. Follow-up (separate issue) if exactness is ever needed: composite `(ts, id)` cursor.
2. `src/api/router.ts:2053` — `order` parsing is case-sensitive and silently falls back to `desc` for any non-`asc` value (`ASC`, `descending`, typos). The safe default is intentional and fine for an internal API; a stricter 400 (or normalize to lowercase) would aid client debugging. Low priority.

✅ **AC Coverage (#211)**
- [x] `PaginationOpts.order?: 'asc'|'desc'`, default `desc`, backward-compatible
- [x] `order='asc'` orders `ts ASC`; `after`+`limit` returns the forward window in one round-trip
- [x] `nextCursor` = last (newest) `ts` for `asc`
- [x] `db.ts` / `types.ts` / `router.ts` / `API.md` all updated
- [~] Test location: issue suggested `tests/unit/api`; placed in `tests/unit/history-db.test.ts` instead because the ordering logic lives in `db.ts` — **documented deviation** in the PR body.

💡 **Suggestions**
- Consider filing the composite-cursor exactness follow-up so the `ts`-tie edge is tracked rather than latent.

🏁 **Verdict: ✅ Ready to merge** — score ≥ 8.0, no Critical/Medium; the two Minors are non-blocking (one pre-existing, one intentional safe-default).

PR #217, opened 2026-07-20
feat: image generation via generate_image MCP tool

0xthoth:
## Summary

Adds image generation to the gateway: a new `generate_image` MCP tool plus the plumbing to deliver generated images through chat channels and the web transcript.

## What's included

- **`generate_image` MCP tool** (`mcp/tools/image/module.ts`) — submits a generation job to an OpenAI-compatible image endpoint, polls until done, and delivers the resulting file. The endpoint defaults to the LLM provider (`ANTHROPIC_BASE_URL`) but can be pointed at a **separate image provider** via `IMAGE_BASE_URL` / `IMAGE_API_KEY`. Hardened: https-only transport for the bearer secret, an SSRF guard on remote image downloads (DNS-resolved private/link-local/cloud-metadata address blocking incl. IPv4-mapped IPv6, `redirect: 'error'`), size-capped downloads (content-length + streaming), a bounded poll loop with per-request timeouts, and filename sanitization.
- **Channel behavior** — channel agents are guided to call `generate_image` instead of describing app-install / API-key setup, to never hand-draw images (SVG / ASCII / HTML), to prefer BYOK models with retry-on-failure, and to send a generated image exactly once. Discord gains a per-session dedup guard against retry-driven resends (marks a file delivered only *after* a successful send, so a transient failure still allows re-delivery).
- **Web transcript** — channel-generated images are captured as message media so they render in the web history, matching the API path.
- **Session meta** — the composer's image config is persisted to session meta (`imageConfig`, surfaced on the sessions list) and accepted on the send endpoint as `image_params`.
- **Media URLs** are signed with the gateway API key (removing a separate signing secret); only raster images are served inline.

## Also included (adjacent hardening)

Two security changes ride along because the image tool reuses the M2M proxy secret and API sessions can now reach more tools:

- **Secret non-disclosure** system note for API sessions (`src/agent/runner.ts`) — instructs the agent to refuse requests to reveal env vars / tokens / `.env` contents.
- **`--disallowedTools` enforcement** for `allow_tools:false` API agents (`src/session/process.ts`) — makes that flag a real capability boundary (built-in Bash/WebFetch/... can no longer exfil owner secrets), not just a prompt hint.

## Config

| Env | Purpose |
|-----|---------|
| `IMAGE_BASE_URL` | Image endpoint override (falls back to `ANTHROPIC_BASE_URL`) |
| `IMAGE_API_KEY` | Image secret override (falls back to `ANTHROPIC_AUTH_TOKEN`) |
| `IMAGE_DISABLED` | Set `true` to disable the tool |
| `IMAGE_POLL_TIMEOUT_MS` | Local poll budget before handing back a `task_id` (default 150000) |

## Tests

New unit suites: `image-module`, `channel-instructions`, `channel-reply-media`, `discord-reply-dedup`, plus extended `session-process` coverage — all green. `tsc --noEmit` is clean for both `src/` and `mcp/`. `API.md` documents `image_params` and `imageConfig`.

## Notes

Rebased onto current `main` (v1.3.32).


0xMaxMa:
**Review note — two intentional design decisions (not defects), for whoever merges:**

1. **`IMAGE_*` env deliberately diverges from `GETPOD_BROWSER_*`.** The browser tool calls a specific `getpod_browser` backend, so its `GETPOD_BROWSER_*` naming is accurate. Image generation, by contrast, targets *any* OpenAI-compatible provider (not necessarily "getpod"), so the vars are named provider-neutral: `IMAGE_BASE_URL` / `IMAGE_API_KEY` / `IMAGE_DISABLED` / `IMAGE_POLL_TIMEOUT_MS`. The inconsistency with the browser module is intentional.

2. **`IMAGE_BASE_URL` reintroduces an image-endpoint override** that an earlier iteration had removed (the "`GETPOD_IMAGE_URL` fallback removed" tests). It is back on purpose: `baseUrl()` resolves `IMAGE_BASE_URL || ANTHROPIC_BASE_URL` and `authToken()` resolves `IMAGE_API_KEY || ANTHROPIC_AUTH_TOKEN`, so an operator can point image generation at a provider **separate from the LLM** while still defaulting to the shared provider. The `baseUrlIsSecure` guard applies to the overridden URL too (http-to-public → disabled + warn), and both vars are documented in `.env.example` and the PR body.

PR #220, opened 2026-07-21
feat(history): raise message-history MAX_LIMIT to 1000 for single-call jump-to-date

yuuyui:
## Summary

Raise the chat-history API page ceiling from **200 to 1000** so a client can fetch a whole `target day → now` span in a **single** request (getpod jump-to-date, Crown-Labs/getpod#1798), instead of several sequential `load-more` pages that read as inter-page "await" flicker. 1000 covers a realistic single-session span (~722 messages for a 3-month mock) with headroom while still bounding the response payload.

Refs Crown-Labs/getpod#1798.

## What changed

- **`src/history/db.ts`** — introduce exported `MAX_HISTORY_LIMIT = 1000`; `MAX_LIMIT` now references it. `DEFAULT_LIMIT` stays `50` (no behavior change for clients that don't request a large limit); `hasMore`/cursor logic untouched.
- **`src/api/router.ts`** — the `GET …/messages` HTTP-boundary clamp now reuses `MAX_HISTORY_LIMIT` instead of a separate literal `200`, so the two ceilings can never drift. **This was a second, independent 200-cap** the issue's implementation notes did not mention: without it, a `limit=1000` request was silently re-clamped to 200 at the HTTP layer before ever reaching the db.

## Incidental hardening (beyond the issue's ACs)

While adding edge-case tests we found a **pre-existing** defect: `getMessages()` never validated `opts.limit`, so a `NaN` or fractional value (e.g. `50.7`) reached the SQLite bind (`limit + 1`) as a non-integer and threw a raw `datatype mismatch`. This is **unreachable via HTTP** (the router's `parseInt` guarantees an integer or a `50` fallback), but `getMessages` is an exported method, so a future direct caller (MCP tool, cron, …) could crash.

Fix: `getMessages()` now integer-coerces `limit` (`NaN → DEFAULT_LIMIT`, fractional → truncated) before the bind — mirroring the HTTP boundary's own `parseInt` behavior. `0`/negative are preserved (explicit and already bounded downstream). This is called out explicitly as it is **outside the stated ACs of #1798**.

## Acceptance criteria (Crown-Labs/getpod#1798)

- [x] `MAX_LIMIT` raised 200 → 1000.
- [x] A request with `limit=1000` returns >200 rows when they exist (integration `I-HIST-30`).
- [x] `DEFAULT_LIMIT` unchanged (50) — no behavior change without a large limit.
- [x] `hasMore`/cursor semantics correct at the new ceiling (still fetches `limit + 1`).
- [ ] New gateway version published to npm + deployed to dev pods — **release/ops step, not in this PR.**
- [ ] getpod client single-call wiring — tracked in Crown-Labs/getpod#510 / PR #1707.

## Testing

- New unit coverage (`tests/unit/history-db.test.ts`): exported ceiling = 1000, >200 rows in one call, unchanged default, clamp-at-ceiling + `hasMore` boundary, and limit input validation (mid-range, below-default, `0`, negative no-leak, `NaN`, fractional, `Infinity`, `MAX_SAFE_INTEGER`).
- New integration coverage (`tests/integration/chat-history-api.test.ts`, `I-HIST-30/31/32`): real `GET …/messages?limit=1000` through the router returns >200 rows; default returns 50; malformed limits are handled cleanly.
- Full suite: **130/130 suites, 2503/2503 tests pass**; typecheck clean.

## Out of scope

- npm republish + pod redeploy (release/ops).
- getpod client wiring (Crown-Labs/getpod#510 / PR #1707).
- Spans > 1000 messages — still fall back to multi-page paging (client loop already handles it).

### DIFF
--- a/API.md
+++ b/API.md
@@ -2377,7 +2377,7 @@ Paginated message history (cursor-based). Returns messages in reverse chronologi
 
 | Param | Description |
 |-------|-------------|
-| `limit` | Max messages to return (default 50, max 200) |
+| `limit` | Max messages to return (default 50, max 1000). Values above the max are clamped; non-numeric, `0`, or negative fall back to the default. |
 | `before` | Return messages before this timestamp (ms) |
 | `after` | Return messages after this timestamp (ms) |
 | `before_id` | Id component of the cursor, paired with `before`. Echo back `nextCursorId` here to page correctly across messages that share a `ts` (see below). Ignored without `before`. |
--- a/src/api/router.ts
+++ b/src/api/router.ts
@@ -9,7 +9,7 @@ import { AgentRunner } from '../agent/runner';
 import { AgentConfig, ApiKey, ImageParams, ModelConfig, SessionMeta } from '../types';
 import { createApiAuthMiddleware, canAccessAgent, canWriteAgent, isAdmin } from './auth';
 import { MediaStore } from '../history/media-store';
-import { HistoryDB } from '../history/db';
+import { HistoryDB, MAX_HISTORY_LIMIT } from '../history/db';
 import type { AgentSessionSummary } from '../history/types';
 import { wizardStore } from './wizard-state';
 import { getPendingSenders, clearPendingSender } from './line-pending-senders';
@@ -2082,7 +2082,9 @@ export function createApiRouter(
       return;
     }
     const query = req.query as Record<string, string>;
-    const limit = query['limit'] ? Math.min(parseInt(query['limit'], 10) || 50, 200) : 50;
+    // Ceiling reuses MAX_HISTORY_LIMIT from the history layer so this HTTP-boundary
+    // clamp and the db clamp can never drift (both 1000; see #1798).
+    const limit = query['limit'] ? Math.min(parseInt(query['limit'], 10) || 50, MAX_HISTORY_LIMIT) : 50;
     // Numeric cursor params. before/after are ms timestamps; before_id/after_id are the id
     // component of the composite (ts, id) cursor, echoed from a prior page's nextCursorId —
     // paired with before/after they stop paging from skipping messages that share a ts
--- a/src/history/db.ts
+++ b/src/history/db.ts
@@ -23,7 +23,6 @@ import {
  * by the HTTP boundary clamp (api/router.ts) so the two ceilings can never drift.
  */
 export const MAX_HISTORY_LIMIT = 1000;
-const MAX_LIMIT = MAX_HISTORY_LIMIT;
 const DEFAULT_LIMIT = 50;
 const PREVIEW_LENGTH = 120;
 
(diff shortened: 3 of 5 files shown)


yuuyui:
## 🔍 Code Review — self-review (updated after addressing findings)

**📊 Score: 8.5 → 9.2 / 10.0** (after the follow-up commit)
- Code Quality: 9.0
- Correctness: 9.5
- Production Readiness: 9.0

### 💪 Strengths
- Kills the root bug-class: the two independent `200` caps (history `db.ts`, `api/router.ts`) are consolidated into a single exported `MAX_HISTORY_LIMIT`, so they can no longer drift.
- Comments explain **why**, not what (drift-prevention on the router clamp; the SQLite `datatype mismatch` rationale on the coercion).
- Tests exercise real behavior through the router via `supertest` (not mocks), with a thorough edge-case set (NaN, ±Infinity, fractional, `MAX_SAFE_INTEGER`, 0, negative-no-leak, duplicate param, unknown chat) and exact-ceiling `hasMore` boundaries at both the db and HTTP layers.

1. 🟢 `tests/unit/history-db.test.ts` — a small `seed(db, n)` helper is defined in two adjacent `describe` blocks. Not addressed here (pure test-local polish); optional follow-up to hoist one shared helper.

### 🔴 Critical (0) — none
### 🟡 Medium (0) — none

### 🟢 Minor — all resolved in the follow-up commit
1. ✅ **Non-finite limit guard** — `getMessages()` guarded only `NaN`, letting `-Infinity` reach the SQLite bind (the exact `datatype mismatch` the fix targets; unreachable via HTTP, but the method is exported). Now uses `Number.isFinite(rawLimit) ? Math.trunc(rawLimit) : DEFAULT_LIMIT`, covering NaN/+Inf/−Inf uniformly. Locked with a `-Infinity` unit test.
2. ✅ **Router-clamp coverage** — added integration `I-HIST-33`: seeds past the ceiling and requests an over-ceiling `limit`, asserting the **HTTP boundary** clamps to exactly `MAX_HISTORY_LIMIT` (distinct from the db-layer clamp the unit tests cover).
3. ✅ **Alias removed** — dropped the `MAX_LIMIT` alias in favor of referencing `MAX_HISTORY_LIMIT` directly (one name for one value); updated a stale comment.

### ✅ AC Coverage (Crown-Labs/getpod#1798)
- [x] AC1 `MAX_LIMIT` 200 → 1000
- [x] AC2 `limit=1000` returns >200 rows in one call (I-HIST-30)
- [x] AC3 `DEFAULT_LIMIT` unchanged (50) — no behavior change without a large limit
- [x] AC4 `hasMore`/cursor correct at the new ceiling
- [ ] ⚠️ AC5 npm publish + pod redeploy — release/ops, out of this PR
- [ ] ⚠️ AC6 getpod client single-call wiring — tracked Crown-Labs/getpod#510 / PR #1707

No Critical/Medium; the three Minors are resolved. Full suite green (the one flaky integration suite passes in isolation and is unrelated to this change).

<sub>Self-review via /wtb-code-review.</sub>

PR #227, opened 2026-07-22
[Feature] Support per-job timezone for recurring cron jobs

### LINKED ISSUE #225: [Enhancement] Support per-job timezone for recurring cron jobs (IANA tz → node-cron)
## Summary
Recurring cron jobs always fire in UTC regardless of the user's timezone. Add an optional per-job `timezone` (IANA) that the scheduler honors, so a job scheduled for "09:00" fires at 09:00 in the user's zone — not 09:00 UTC.

**Priority:** High

## Background / Context
The gateway process runs in UTC (`config.json` `timezone: "UTC"`). In `src/cron/manager.ts`, `scheduleJob()` schedules recurring jobs with:

```ts
const task = nodeCron.schedule(job.schedule!, async () => { await this.executeJob(job); });
```

No options object is passed, so `node-cron` interprets the cron expression in the **process** timezone (UTC). A user in `Asia/Bangkok` who schedules `0 9 * * *` ("09:00") actually gets a job that fires at **16:00** their local time, and the getpod frontend cron preview (`crons-panel.tsx:45`) is off by the same offset.

One-shot `at` jobs are **already correct** — `scheduleJob()` handles them via `Date.parse(job.scheduleAt!)` on an absolute ISO-8601 instant, which carries no zone ambiguity. This issue is scoped to `kind='cron'` recurring jobs only.

## Acceptance Criteria
- [ ] `CronJob`, `CronJobCreate`, and `CronJobUpdate` (`src/types.ts:297,320,339`) gain an optional `timezone?: string` (IANA zone name, e.g. `"Asia/Bangkok"`).
- [ ] For `kind='cron'`, `scheduleJob()` passes `{ timezone: job.timezone ?? 'UTC' }` as the third argument to `nodeCron.schedule(...)`, so the job fires in the stored zone (DST-safe via node-cron).
- [ ] `create()` and `update()` persist `timezone`; an invalid / unresolvable IANA zone is rejected with a clear validation error (alongside the existing `validateSchedule` checks).
- [ ] The cron API (`src/api/gateway-router.ts`) accepts `timezone` on create/update and returns it in `list`/`get` responses.
- [ ] Existing/legacy jobs with no `timezone` default to `'UTC'` → identical behavior to today (backward compatible).
- [ ] Unit tests: (a) a job with `timezone` set is scheduled in that zone; (b) a job with no `timezone` defaults to UTC.

## Out of Scope
- One-shot `at` jobs (already timezone-correct via absolute instant).
- The heartbeat scheduler (`src/cron/scheduler.ts`) — separate code path, not user-facing cron.
- Frontend UI work (timezone detection, preview fix, presets) — tracked in getpod frontend issue #1690.

## How to Implement
1. **Types** — `src/types.ts`: add `timezone?: string` to `CronJob` (~:297), `CronJobCreate` (~:320), and `CronJobUpdate` (~:339).
2. **Scheduler** — `src/cron/manager.ts` `scheduleJob()`, `kind='cron'` branch:
   ```ts
   const task = nodeCron.schedule(
     job.schedule!,
     async () => { await this.executeJob(job); },
     { timezone: job.timezone ?? 'UTC' },
   );
   ```
3. **Persist + validate** — `create()` (~:129) and `update()` (~:176): copy `input.timezone` onto the job; extend `validateSchedule()` (~:289) to reject an unresolvable zone (e.g. probe with `Intl.DateTimeFormat(undefined, { timeZone })` in a try/catch).
4. **API** — `src/api/gateway-router.ts`: thread `timezone` through the create/update request parsing and the job serialization.
5. **Tests** — extend the cron manager test suite: assert the third-arg options carry the expected `timezone`, and that a missing `timezone` yields `'UTC'`.

## Dependencies
- Related to: getpod frontend #1690 (recurring cron presets + fire/show in the user's timezone) — **this issue is its gateway-side prerequisite**. The frontend cannot fire jobs in the user's local zone until the gateway stores and honors a per-job timezone.
- Blocked by: none.

## Notes
- Doing the local→UTC offset conversion purely on the frontend would break across DST boundaries (a job could fire an hour off, or on the wrong day for weekly/monthly schedules). Storing the IANA zone and letting `node-cron` resolve it at fire time is the DST-safe design.
- `node-cron` supports the `{ timezone }` option and resolves DST internally — no manual offset math required.
- After merge: the gateway is a separately published npm package (`@0xmaxma/claude-gateway`), so a preview environment only picks this up after republish + reinstall + restart.

yuuyui:
## Summary
Adds an optional per-job `timezone` field to recurring cron jobs so a scheduled job fires at the intended wall-clock time in a specific IANA zone (e.g. `Asia/Bangkok`), instead of always being interpreted in the scheduler's process zone (UTC). Both the live schedule and the startup catch-up path are now zone-aware, and invalid zones are rejected at create/update time.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Enhancement
- [ ] Refactor / Tech debt
- [ ] Documentation

## Related Issues
Closes #225

## Changes Made
- `src/types.ts`: add optional `timezone?: string` to the recurring cron job types.
- `src/cron/manager.ts`:
  - `scheduleJob()` now passes `{ timezone: job.timezone ?? 'UTC' }` to `nodeCron.schedule(...)` so live scheduling is zone-aware.
  - `catchUpMissedJobs()` now forwards `tz: job.timezone ?? 'UTC'` to `CronExpressionParser.parse(...)` so restart catch-up computes the last-expected tick in the job's own zone (previously used the process zone — a Bangkok job could be caught up ~7h off).
  - `create()` / `update()` persist and re-apply `timezone`; `validateSchedule()` rejects an unresolvable zone via `Intl.DateTimeFormat` and rejects empty/whitespace/non-string values without silently overwriting an existing value.
- `tests/unit/cron-manager.test.ts`: 15 timezone cases (TZ1–TZ15) covering valid IANA/lowercase/fixed-offset zones, empty/whitespace/null/number rejection, at-job passthrough, multi-job isolation, and the restart catch-up path (TZ-GAP).

## How to Test
1. Checkout this branch and run the cron suite under a non-UTC zone: `TZ=Asia/Bangkok npx jest cron-manager`.
2. Create a recurring job with `timezone: "Asia/Bangkok"` and a daily schedule; confirm it fires at the Bangkok wall-clock time, not UTC.
3. Restart the gateway shortly after a job's scheduled time and confirm catch-up does not double-fire or skip.

Expected: all cron tests green (79/79); zone-aware jobs fire at the correct local time; invalid zones are rejected at create/update.

## Checklist
- [x] Reviewed my own code
- [x] Tests pass (if applicable)
- [x] No console errors

### DIFF
--- a/API.md
+++ b/API.md
@@ -1913,6 +1913,7 @@ Jobs are persisted to `~/.claude-gateway/crons.json` and survive gateway restart
 | `scheduleKind` | No | `"cron"` (default) or `"at"` |
 | `schedule` | If `scheduleKind=cron` | 5-field cron expression e.g. `"0 9 * * *"` |
 | `scheduleAt` | If `scheduleKind=at` | ISO 8601 timestamp for one-shot run |
+| `timezone` | No | IANA zone (e.g. `"Asia/Bangkok"`) the `scheduleKind=cron` expression fires in — DST-safe. Defaults to `"UTC"` (legacy jobs unchanged). An unresolvable zone is rejected with `400`. Ignored for `scheduleKind=at` (absolute instants carry no zone ambiguity). |
 | `type` | No | `"command"` (default) or `"agent"` |
 | `command` | If `type=command` | Shell command to run |
 | `prompt` | If `type=agent` | Prompt sent to the agent as a new turn |
@@ -2006,6 +2007,26 @@ curl -s -X POST http://localhost:10850/api/v1/crons \
     "name": "morning-brief",
     "scheduleKind": "cron",
     "schedule": "0 9 * * *",
+    "type": "agent",
+    "prompt": "Give me a morning summary.",
+    "telegram": "<CHAT_ID>"
+  }' | jq
+```
+
+#### Example: Daily job in a specific timezone
+
+Run every day at 09:00 **Bangkok time** — not 09:00 UTC. Add `timezone` (any IANA zone); node-cron resolves DST at fire time.
+
+```bash
+curl -s -X POST http://localhost:10850/api/v1/crons \
+  -H "X-Api-Key: my-secret-key-123" \
+  -H "Content-Type: application/json" \
+  -d '{
+    "agentId": "claude-founder",
+    "name": "morning-brief-bkk",
+    "scheduleKind": "cron",
+    "schedule": "0 9 * * *",
+    "timezone": "Asia/Bangkok",
     "type": "agent",
     "prompt": "Give me a morning summary.",
     "telegram": "<CHAT_ID>"
--- a/src/cron/manager.ts
+++ b/src/cron/manager.ts
@@ -198,6 +198,10 @@ export class CronManager extends EventEmitter {
     if (input.scheduleKind !== undefined) job.scheduleKind = input.scheduleKind;
     if (input.schedule !== undefined) job.schedule = input.schedule;
     if (input.scheduleAt !== undefined) job.scheduleAt = input.scheduleAt;
+    // Once set, a timezone can be changed to another valid zone but not cleared
+    // back to "unset": `undefined` is the no-op guard above and `null` is rejected
+    // by validateTimezone. To restore the default behavior, set it to "UTC"
+    // explicitly — which is identical to an unset zone (see scheduleJob()'s ?? 'UTC').
     if (input.timezone !== undefined) job.timezone = input.timezone;
     if (input.type !== undefined) job.type = input.type;
     if (input.command !== undefined) job.command = input.command;
--- a/src/cron/manager.ts
+++ b/src/cron/manager.ts
@@ -621,7 +621,13 @@ export class CronManager extends EventEmitter {
       // more than 30 minutes ago, regardless of whether the schedule produced a new tick.
       let lastExpectedRun: Date;
       try {
-        const interval = CronExpressionParser.parse(job.schedule!, { currentDate: new Date() });
+        // Interpret the schedule in the job's own timezone (matches scheduleJob()'s
+        // node-cron {timezone} default) so catch-up computes the correct missed tick
+        // for non-UTC jobs after a restart.
+        const interval = CronExpressionParser.parse(job.schedule!, {
+          currentDate: new Date(),
+          tz: job.timezone ?? 'UTC',
+        });
         lastExpectedRun = interval.prev().toDate();
       } catch {
         continue; // unparseable schedule — skip catch-up
--- a/src/cron/manager.ts
+++ b/src/cron/manager.ts
@@ -141,6 +141,7 @@ export class CronManager extends EventEmitter {
       scheduleKind,
       schedule: input.schedule,
       scheduleAt: input.scheduleAt,
+      timezone: input.timezone,
       type,
       command: input.command,
       prompt: input.prompt,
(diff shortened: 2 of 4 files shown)


yuuyui:
## 🔍 Code Review — Self-review

**📊 Score: 9.0 / 10.0**

| Dimension | Score |
|---|---|
| Code Quality | 9.0 |
| Correctness | 9.0 |
| Production Readiness | 9.0 |

### 💪 Strengths
- **`catchUpMissedJobs` zone-skew gap caught & fixed** (`10cf989` → rebased `ac6c261`) — the non-obvious *second* zone-sensitive path (restart catch-up) beyond live scheduling. cron-parser's `tz` option is correctly distinguished from node-cron's `timezone` (an easy trap).
- **Tests exercise real behavior, not mock echo** — they spy on the real `node-cron` / `cron-parser` modules and assert the actual options object handed to `schedule()` / `parse()` (TZ1–15 + TZ-GAP).
- **Backward compatibility preserved** — `?? 'UTC'` reproduces the prior no-options UTC behavior exactly for legacy jobs (locked in by TZ2).
- **Intent-revealing comments** at every zone-sensitive site; `validateTimezone` probes with `Intl` (the same machinery node-cron resolves through) rather than a hand-rolled zone list.

1. 🟢 `manager.ts` — the `?? 'UTC'` default literal appears in 3 spots (schedule options, log line, catch-up parse). Not logic duplication, but a magic default worth a single `const DEFAULT_CRON_TZ = 'UTC'` to keep the three in lockstep. *(non-blocking)*

### 🔴 Critical (0)
_none_

### 🟡 Medium (0)
_none_

### 🟢 Minor (2) — both addressed in `7a1da29`
1. **`+07:00` fixed-offset not scheduled end-to-end** — TZ9 originally asserted only that `create()` accepts the offset, not that `scheduleJob` fires with it. → **Fixed:** TZ9 now asserts `nodeCron.schedule` receives `{ timezone: '+07:00' }`.
2. **`timezone` cannot be cleared back to unset** — once set it can only change to another valid zone (`undefined` is a no-op guard, `null` rejected by TZ12). → **Documented** in `update()`: set `"UTC"` explicitly for the equivalent default. Acceptable by design.

### ✅ AC Coverage (issue #225)
- [x] **AC1** — `timezone?` on `CronJob` / `CronJobCreate` / `CronJobUpdate` (`types.ts`)
- [x] **AC2** — `scheduleJob` passes `{ timezone: job.timezone ?? 'UTC' }` (`manager.ts`)
- [x] **AC3** — create/update persist + reject invalid zone (`validateTimezone`, TZ3/TZ10/TZ15)
- [x] **AC4** — API accepts `timezone` + returns it — met via `cron-router.ts` whole-body pass-through (create/update spread `req.body`; list/get serialize the job). *Note: issue named `gateway-router.ts`, but the routes live in `cron-router.ts` — no explicit router change needed. Verified, not skipped.*
- [x] **AC5** — legacy jobs default to UTC, backward compatible (TZ2)
- [x] **AC6** — unit tests (a) zoned job (b) UTC default — plus 14 more (TZ3–15, TZ-GAP)

**Verification:** `TZ=Asia/Bangkok` → cron-manager 39/39, full cron suite 79/79, whole project 2533/2533, `tsc --noEmit` clean. Independently re-run by tester on all commits.

---

PR #231, opened 2026-07-23
fix(security): require auth for dashboard/status/processes on non-loopback bind

### LINKED ISSUE #230: [Security] Dashboard/status/processes reachable unauthenticated when gateway.bind is non-loopback
## Summary

When `gateway.bind` is set to a non-loopback interface (e.g. `0.0.0.0`), several HTTP endpoints are reachable **without any authentication**, because auth is not applied globally — it lives only inside the `/api/*` sub-routers. Everything mounted directly on the app (health, dashboard, status, processes) is gated **only** by the localhost bind. The config migrator can pin `bind = 0.0.0.0` on upgrade (behavior-preserving for the pre-1.0.13 default), so real deployments end up exposed.

## Investigation (grounded in `src/api/gateway-router.ts`)

Unauthenticated surface today (with `bind = 0.0.0.0`):

- `GET /health` (`:476`) — returns the full agent-id list (`{status, agents:[...]}`).
- `GET /status` (`:595`) — per-agent stats **including `sessionId`** (via `getSessionsSummary()`), message counts, heartbeat history.
- `GET /processes` (`:502`) — host `ps` output including full command-line args of every gateway/claude/mcp process.
- `GET /dashboard` (`:481`) — serves the dashboard and mints a `dashToken` with **no auth**.

The dashboard token is the dangerous link: `GET /dashboard` (unauth) → `dashToken` → `POST /api/v1/pty-stream-ticket` (accepts the token) → WS `/api/v1/agents/:id/pty-stream` → **live PTY stream that also routes inbound frames as keystrokes into the agent's shell** (`attachPtyStreamSocket` → `sendInputToSession`). `sessionId` needed for the ticket is itself disclosed by the unauthenticated `/status`. Agents run with `--dangerously-skip-permissions`, so unauthenticated keystroke injection is effectively host RCE on any PTY-mode (non-headless) session. The code comment at `attachPtyStreamSocket` states access is "protected upstream ... and the gateway binds to localhost by default" — i.e. the localhost bind is the sole boundary, which `bind = 0.0.0.0` removes.

The `/api/*` control API (send message, agent CRUD, workspace files, skills, cron, apps) **is** properly authenticated (`router.ts` `createApiAuthMiddleware` + `canAccessAgent`/`canWriteAgent`); LINE webhooks verify HMAC. Those are out of scope. The MCP callback server is hardcoded to `127.0.0.1`; the app/PTY control sockets are Unix-domain — none are affected by `gateway.bind`.

## Proposed approach

Add authentication to the dashboard-adjacent surface, keyed on the existing `gateway.api.keys` (auth stays disabled when no keys are configured, to avoid locking out keyless local installs):

1. **`/health`** → return only `{"status":"ok"}` (drop the agent-id list). Stays public for liveness probes but leaks nothing.
2. **`/status` + `/processes`** → require a valid API key (`X-Api-Key`/`Bearer`) **or** a dashboard session; `401` otherwise.
3. **Dashboard session via `HttpOnly` cookie:**
   - `GET /dashboard` serves a login page (API-key form) when there is no valid `dash_session` cookie, else the dashboard.
   - `POST /dashboard/login` validates an API key (timing-safe) and sets `dash_session` (`HttpOnly; SameSite=Strict; Secure` when TLS; ~8h, multi-use).
   - `POST /dashboard/logout` revokes it.
   - Gate `/status`, `/processes`, `/api/v1/sessions/:id/screen`, and `POST /api/v1/pty-stream-ticket` on API key **or** cookie.
   - Stop embedding any token in the page (`<meta name="dash-token">` removed) — the browser sends the cookie automatically; the ticket exchange keeps credentials out of the WS URL. This also lets the session be multi-use (the old one-time-use hack existed only because the token was readable via view-source).

## Acceptance criteria

- [ ] `GET /health` returns exactly `{"status":"ok"}` — no agent ids or version.
- [ ] With `gateway.api.keys` configured: `GET /status` and `GET /processes` return `401` without a credential, `200` with a valid API key, and `200` with a valid dashboard session cookie.
- [ ] `GET /dashboard` serves a login page when unauthenticated; `POST /dashboard/login` with a valid key sets an `HttpOnly; SameSite=Strict` cookie; a wrong key returns `401` with no cookie.
- [ ] The dashboard no longer embeds any auth token in the HTML (`name="dash-token"` gone); the live-terminal chain requires a valid session.
- [ ] `POST /dashboard/logout` revokes the session (the same cookie no longer authorizes `/status`).
- [ ] Keyless installs (no `gateway.api.keys`) keep the dashboard and monitoring endpoints open (no lock-out).
- [ ] Authorization is not weakened elsewhere; existing broken/exposed state is closed without manual intervention.
- [ ] Regression tests cover the above and fail on the pre-fix code; `API.md`/`README.md` updated.

0xMaxMa:
Closes #230

## What & why

With `gateway.bind` set to a non-loopback interface (`0.0.0.0`), the dashboard-adjacent endpoints were reachable **without any credential** — auth is applied only inside the `/api/*` sub-routers, not globally, so everything mounted directly on the app was gated solely by the localhost bind. The unauthenticated `/dashboard` minted a token that could be exchanged at `POST /api/v1/pty-stream-ticket` for a live PTY stream, which also routes inbound frames as **keystrokes into the agent shell** — effectively unauthenticated RCE on PTY-mode sessions (agents run with `--dangerously-skip-permissions`). `/status` disclosed the `sessionId` that chain needs, `/health` leaked the agent-id list, and `/processes` leaked host command-line args.

## Changes

Auth is gated on the existing `gateway.api.keys` — when **no** keys are configured it stays open, so keyless local installs are not locked out (mirrors how the `/api` routers are only mounted when keys exist).

- **`/health`** → returns only `{"status":"ok"}` (no agent ids). Still public for liveness probes.
- **`/status` + `/processes`** → require an API key (`X-Api-Key`/`Bearer`) **or** a dashboard session cookie; `401` otherwise.
- **Dashboard session via `HttpOnly` cookie:**
  - `GET /dashboard` serves a login page (API-key form) when there is no valid `dash_session` cookie, else the dashboard.
  - `POST /dashboard/login` validates a key (timing-safe) and sets `dash_session` (`HttpOnly; SameSite=Strict; Path=/`; `Secure` when TLS; 8h; multi-use).
  - `POST /dashboard/logout` revokes the session and clears the cookie.
  - `/status`, `/processes`, `/api/v1/sessions/:id/screen`, and `POST /api/v1/pty-stream-ticket` accept the API key **or** the cookie.
  - The page no longer embeds any auth token (`<meta name="dash-token">` removed); the browser sends the cookie automatically and the ticket exchange keeps credentials out of the WS URL. The client redirects to the login page on `401` and gains a **Logout** button.
- **WS header auth** now uses a timing-safe comparison, matching the `/api` middleware.

Cookies are parsed with a small manual helper — no new dependency.

## Out of scope

- The `/api/*` control API is already authenticated (`createApiAuthMiddleware` + `canAccessAgent`/`canWriteAgent`); LINE webhooks verify HMAC. Unchanged.
- The `/app/:name/:portName/*` reverse proxy remains unauthenticated (each app authenticates itself). Tracked separately — not part of this dashboard-hardening change.

## Testing

- New `tests/unit/gateway-dashboard-auth.test.ts` (15 cases): `/health` minimal, `/status`+`/processes` 401/200 (key + cookie), login/logout/forged-cookie, and the keyless-open path. **Proven to fail on the pre-fix code** (10/15 red when the `src` changes are reverted).
- Updated the existing `/health`-lists-agents assertions (e2e I-E2E-01/07, multi-agent I-MA-09) to read the agent list from `/status`.
- `npm run typecheck` clean; full suite `2532` tests pass (`workspace-loader` is a pre-existing parallel-race flake, green in isolation).
- `API.md` and `README.md` updated (endpoint table, `/health` + `/status` sections, `gateway.bind` + Terminal Viewer security notes).

### DIFF
--- a/API.md
+++ b/API.md
@@ -16,7 +16,7 @@ All API endpoints require an API key configured in `config.json`. Pass it via:
 | `GET` | `/status` | Key or dashboard session¹ | Per-agent stats + heartbeat history |
 | `GET` | `/processes` | Key or dashboard session¹ | Host process tree for the dashboard |
 | `GET` | `/dashboard` | Session cookie¹ | Web UI dashboard (serves the login page when unauthenticated) |
-| `POST` | `/dashboard/login` | None (validates a key) | Exchange an API key for an `HttpOnly` `dash_session` cookie (8h) |
+| `POST` | `/dashboard/login` | None (validates a key) | Exchange an API key for an `HttpOnly; SameSite=Lax` `dash_session` cookie (8h). Brute-force throttled per IP (`429` after 10 failed attempts / 5 min) |
 | `POST` | `/dashboard/logout` | Session cookie | Revoke the dashboard session and clear the cookie |
 | `GET` | `/api/v1/commands` | None | List slash commands available in the chat UI |
 
@@ -213,8 +213,8 @@ each session of an agent is an isolated stream.
 | `POST` | `/api/v1/pty-stream-ticket` | Key *or* dashboard session | Exchange an API key **or** the `dash_session` cookie for a one-time, 30s-TTL ticket bound to a specific `{ agentId, sessionId }` |
 | `WS` | `/api/v1/agents/:agentId/pty-stream` | Ticket *or* Key | Subscribe to the live PTY stream for one session |
 
-> The browser dashboard authenticates with the `HttpOnly` `dash_session` cookie
-> (from `POST /dashboard/login`), which is sent automatically — it no longer embeds
+> The browser dashboard authenticates with the `HttpOnly; SameSite=Lax` `dash_session`
+> cookie (from `POST /dashboard/login`), which is sent automatically — it no longer embeds
 > any token in the page. `POST /api/v1/pty-stream-ticket` accepts that cookie, so
 > the ticket flow works without a token in the HTML or the WS URL.
 
--- a/API.md
+++ b/API.md
@@ -20,11 +20,14 @@ All API endpoints require an API key configured in `config.json`. Pass it via:
 | `POST` | `/dashboard/logout` | Session cookie | Revoke the dashboard session and clear the cookie |
 | `GET` | `/api/v1/commands` | None | List slash commands available in the chat UI |
 
-¹ **Auth applies only when `gateway.api.keys` is configured.** With no keys, these
-endpoints stay open (a keyless install has no credential to check). "Key or dashboard
-session" accepts an API key (`X-Api-Key` / `Authorization: Bearer`) **or** the
-`dash_session` cookie issued by `POST /dashboard/login`. When bound to a non-loopback
-interface (`gateway.bind = 0.0.0.0`), configure API keys so these are not world-readable.
+¹ **Auth applies when `gateway.api.keys` is configured.** "Key or dashboard session"
+accepts an API key (`X-Api-Key` / `Authorization: Bearer`) **or** the `dash_session`
+cookie issued by `POST /dashboard/login`. With **no** keys configured the behavior
+depends on the bind: on a **loopback** bind (`127.0.0.1`) they stay open (a keyless
+local install has no credential to check); on a **non-loopback** bind (`0.0.0.0` or a
+real IP) they **fail closed** — `/status`, `/processes`, and `/dashboard` return `503`
+until `gateway.api.keys` is set, so the surface is never exposed unauthenticated to the
+network. `/health` stays public in all cases.
 
 ### Agent API
 
--- a/API.md
+++ b/API.md
@@ -12,10 +12,19 @@ All API endpoints require an API key configured in `config.json`. Pass it via:
 
 | Method | Path | Auth | Description |
 |--------|------|------|-------------|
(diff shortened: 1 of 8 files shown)


0xMaxMa:
Added a follow-up commit (per review discussion): **fail closed for keyless installs on a non-loopback bind.** Previously `/status`, `/processes`, and `/dashboard` stayed open with no `gateway.api.keys` regardless of the bind — so a keyless `0.0.0.0` deployment was still exposed. Now:

- Keyless + loopback (`127.0.0.1`) → open, as before (no local lock-out).
- Keyless + non-loopback (`0.0.0.0` / real IP) → `/status`, `/processes`, `/dashboard` return `503` until keys are set (startup warning logged). `/health` stays public.

6 regression tests added and verified to fail on the pre-fix code. The `/app/:name/:portName/*` proxy is intentionally left as-is (each app self-authenticates) — out of scope for this dashboard-hardening PR.

PR #254, opened 2026-08-01
[Fix] Allow channel-less agent cron jobs on create (#253)

### LINKED ISSUE #253: [Bug] Cron create rejects channel-less agent jobs despite Run History support
## Summary
The cron **create** path rejects `type=agent` jobs that have no Telegram/Discord target, even though the execution engine already runs such jobs and persists their output to Run History. The channel requirement should be lifted at create time so channel-less ("web/Run-History only") agent jobs can be scheduled.

**Priority:** Medium

## Background / Context
When scheduling an agent cron job from the getpod web UI, the form tells users the channel target is optional — *"leave blank and read the result in Run History."* That copy matches the intended design: the cron executor already runs the agent turn and writes the run log **unconditionally**; delivering the output to Telegram/Discord is an optional extra.

But `POST /v1/crons` (and `CronManager.validatePayload`) **hard-require** `telegram || discord` for `type=agent`, so creating a channel-less agent job returns:

```
400  telegram or discord is required for type=agent
```

The engine supports the behavior; only the create-time gate blocks it. This surfaced during Crown-Labs/getpod #1690 cron testing: creating a new job on an agent connected to no channel is blocked, contradicting both the UI copy and the intended "web-only / Run History" use case.

## Current Behavior (evidence)

**The create-time gate that over-validates (the bug):**
- `src/api/cron-router.ts:107-110` — the POST handler returns 400 when `type === 'agent' && !body.telegram && !body.discord`.
- `src/cron/manager.ts:326-337` — `validatePayload()` throws `telegram or discord is required for type=agent` under the same condition (runs on `create()`, and on `update()` whenever payload fields change).

**Execution already supports channel-less agent jobs (proof the gate is unnecessary):**
- `src/cron/manager.ts` → `executeJob()`: `runAgentTurn(job)` runs regardless of any channel; `appendRunLog(job.id, runLog)` persists the output to Run History **unconditionally** (~line 469).
- Delivery is optional and guarded: `if (job.telegram) deliveries.push(sendTelegram(...))`, `if (job.discord) deliveries.push(sendDiscord(...))`, then `if (deliveries.length > 0) await Promise.allSettled(...)` (~lines 474-482). Zero channels → no delivery attempted, but the run still executes and is logged.

## Acceptance Criteria
- [ ] `POST /v1/crons` with `type=agent`, a valid `prompt`, and **no** `telegram`/`discord` returns `201` and creates the job.
- [ ] The created channel-less agent job executes on schedule and its output is written to the run log (Run History), with **no** delivery attempted.
- [ ] An agent job **with** a channel still delivers to that channel exactly as before (no regression).
- [ ] `prompt` remains required for `type=agent`; `command` remains required for `type=command` — only the channel requirement is lifted.
- [ ] `validatePayload()` no longer throws solely because both channels are empty for `type=agent`.
- [ ] Unit tests cover: channel-less create succeeds; channel-ful create + delivery unchanged; `prompt`-missing still rejected.

## Out of Scope
- **getpod web UI copy** — it already describes the correct (post-fix) behavior; no getpod change is needed once this lands.
- Adding **LINE** as a delivery target (#234).
- Any change to `update()` semantics beyond the payload-validation path it already shares with `create()`.

## How to Implement
1. `src/api/cron-router.ts` (~line 107): remove the `type === 'agent' && !body.telegram && !body.discord → 400` block. Keep the `prompt`-required check directly above it.
2. `src/cron/manager.ts` → `validatePayload()` (~line 333): remove the `if (!input.telegram && !input.discord) throw ...` branch inside the `agent` case. Keep the `prompt` check.
3. Confirm `executeJob()` needs **no** change — it already no-ops delivery when both channels are absent (guarded pushes + `deliveries.length > 0`).
4. Update/extend unit tests for `create()` and `validatePayload()` accordingly.
5. After merge: republish the npm package (`@0xmaxma/claude-gateway`), reinstall globally, and restart the gateway so preview/live environments pick up the change.

## Dependencies
- Blocked by: none
- Related to: #234 (cron delivery targets)

## Notes
- Discovered via **Crown-Labs/getpod #1690** (cron presets) self-testing. The getpod form copy (*"leave blank … read the result in Run History"*) reflects the intended behavior; a getpod-side copy change was considered and then reverted once this root cause was confirmed.
- Minimal-risk change: it only **widens** what `create()` accepts; existing channel-ful jobs are unaffected.
- Line numbers are from `origin/main` at the time of writing.

yuuyui:
## Summary
Cron **create** rejected agent jobs that had no Telegram/Discord destination, even though the run engine fully supports channel-less agent jobs (results are always persisted to Run History; delivery is the only thing that's skipped). This contradicted the getpod form copy that tells users they can leave the channel blank and read the result in Run History. This PR removes the over-strict require-channel guard on create so a channel-less agent job can be created and runs to Run History as designed.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor / Tech debt
- [ ] Documentation

## Related Issues
Closes #253

## Root Cause
The engine already supports channel-less agent jobs:
- `executeJob` → `appendRunLog` persists the run result to Run History **unconditionally**.
- Delivery is guarded by `if (job.telegram)` / `if (job.discord)` — no channel simply means delivery is skipped, the job still runs.

The bug was purely in **create-time validation**, which hard-blocked the job before it could ever run:
- `src/api/cron-router.ts` — returned `400 "telegram or discord is required"` on create.
- `src/cron/manager.ts` — the same check in `validatePayload` (runs on both create and update).

## Changes Made
- `src/api/cron-router.ts`: removed the API-layer require-channel `400` on create.
- `src/cron/manager.ts`: removed the same require-channel check in `validatePayload` (so create **and** update are consistent). The `prompt`-required rule for `type=agent` is unchanged, and channel-delivery logic when a channel *is* present is untouched.
- `tests/unit/cron-router.test.ts`: inverted T23b — channel-less agent create must now return `201` (previously locked the buggy `400`).
- `tests/unit/cron-manager.test.ts`: added a manager-level test (T11b) proving a channel-less agent job runs, persists its output to the run log, and skips delivery; plus tests covering the previously-uncovered update path (`PUT /v1/crons/:id`).

## How to Test
1. Check out this branch.
2. Create an agent cron job on an agent with **no** Telegram/Discord connected (empty channel destination).
3. Expected: creation succeeds (`201`), the job runs on schedule, and its output appears in **Run History** with no delivery attempted.
4. Regression: creating/updating an agent job *with* a channel still delivers as before; `type=agent` still requires a `prompt`.

## Out of Scope
- No change to delivery behavior when a channel is present.
- No change to the `prompt`-required rule for `type=agent`.

## Checklist
- [x] Reviewed my own code
- [x] Tests pass (typecheck ✓ · cron-router + cron-manager 62/62 · full unit suite green)
- [x] No console errors

Discovered via getpod #1690 (cron UX) manual UI testing on `chubby-tunnel.getpod.ai/crons`.

### DIFF
--- a/API.md
+++ b/API.md
@@ -1974,8 +1974,8 @@ Jobs are persisted to `~/.claude-gateway/crons.json` and survive gateway restart
 | `type` | No | `"command"` (default) or `"agent"` |
 | `command` | If `type=command` | Shell command to run |
 | `prompt` | If `type=agent` | Prompt sent to the agent as a new turn |
-| `telegram` | If `type=agent` | Telegram chat_id to deliver the agent response |
-| `discord` | If `type=agent` | Discord channel_id to deliver the agent response |
+| `telegram` | No | Telegram chat_id to deliver the agent response (optional for `type=agent`) |
+| `discord` | No | Discord channel_id to deliver the agent response (optional for `type=agent`) |
 | `timeoutMs` | No | Execution timeout in ms (default 120000) — applies to both `command` and `agent` |
 | `deleteAfterRun` | No | `true` to auto-delete after first run (one-shot jobs) |
 | `enabled` | No | `true` (default) / `false` to create disabled |
@@ -1985,11 +1985,11 @@ Jobs are persisted to `~/.claude-gateway/crons.json` and survive gateway restart
 | | `command` | `agent` |
 |---|---|---|
 | Runs | Shell command | Agent turn (new Claude session) |
-| Key field | `command` | `prompt` + `telegram` and/or `discord` |
+| Key field | `command` | `prompt` (channels optional) |
 | Output | stdout/stderr | Agent response text |
-| Delivery | Logged only | Sent to Telegram and/or Discord |
+| Delivery | Logged only | Sent to Telegram and/or Discord if set, otherwise logged only |
 
-> **Note:** For `type=agent`, at least one of `telegram` or `discord` is required. Both can be set to deliver to multiple channels simultaneously.
+> **Note:** For `type=agent`, only `prompt` is required. `telegram` and `discord` are optional — set either (or both) to deliver the agent response to those channels; with neither, the job still runs on schedule and its response is logged only (no delivery).
 
 ---
 
--- a/src/api/cron-router.ts
+++ b/src/api/cron-router.ts
@@ -104,10 +104,9 @@ export function createCronRouter(manager: CronManager, apiKeys?: ApiKey[], known
       res.status(400).json({ error: 'prompt is required for type=agent' });
       return;
     }
-    if (type === 'agent' && !body.telegram && !body.discord) {
-      res.status(400).json({ error: 'telegram or discord is required for type=agent' });
-      return;
-    }
+    // Note: telegram/discord are optional for type=agent. A channel-less job still
+    // runs and persists its result to Run History (executeJob → appendRunLog is
+    // unconditional; delivery is skipped when no channel is set). See #253.
 
     try {
       const job = await manager.create(body as CronJobCreate);
--- a/src/cron/manager.ts
+++ b/src/cron/manager.ts
@@ -329,9 +329,8 @@ export class CronManager extends EventEmitter {
       if (!input.command) throw new Error('command is required for type=command');
     } else if (kind === 'agent') {
       if (!input.prompt) throw new Error('prompt is required for type=agent');
-      if (!input.telegram && !input.discord) {
-        throw new Error('telegram or discord is required for type=agent');
-      }
+      // telegram/discord are optional: a channel-less agent job runs and its
+      // output is saved to Run History; delivery is simply skipped (see #253).
     }
   }
 
--- a/tests/unit/cron-manager.test.ts
+++ b/tests/unit/cron-manager.test.ts
@@ -400,6 +400,75 @@ describe('T11-T13: Telegram delivery', () => {
     manager.stop();
   });
 
+  it('T11c: update() on an existing channel-less agent job → prompt change succeeds, not rejected (#253)', async () => {
+    const { manager, agentId } = makeManager();
+    await manager.start();
+
+    const job = await manager.create({
+      agentId,
+      name: 'no-channel-update',
+      scheduleKind: 'cron',
+      schedule: '* * * * *',
+      type: 'agent',
+      prompt: 'hello',
+      // no telegram, no discord
+    });
+
+    // Changing `prompt` triggers manager.update()'s validatePayload(merged) branch.
+    // Before #253, merged still had no telegram/discord and this threw
+    // "telegram or discord is required for type=agent" even though the update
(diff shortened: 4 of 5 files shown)


yuuyui:
## 🔍 Code Review — PR #254 `[Fix] Allow channel-less agent cron jobs on create` (self-review)

**📊 Score: 9.5 / 10.0**
- Code Quality:         9.5
- Correctness:          9.5
- Production Readiness: 9.5

### 💪 Strengths
- **Root cause correctly localized to validation, not the engine.** The fix removes the over-strict require-channel guard at *both* layers — `cron-router.ts` (API boundary) and `manager.ts` `validatePayload` (core, runs on create + update) — so the update path becomes consistent for free, not just create.
- **Tests verify real behavior, not mocks.** `T11b` asserts both sides of the claim: the run actually executes and persists output (`log.status === 'ok'`, output captured → Run History) **and** no channel delivery fires (`/sendMessage` fetch call is `undefined`). Positive + negative assertion.
- **Proper regression guard.** `T11e` proves the fix did *not* loosen the `prompt`-required rule for `type=agent` — a common over-correction this explicitly prevents. `T11c`/`T11d` also close the previously-uncovered update path (`PUT /v1/crons/:id`).
- **Deleted code + explanatory comment referencing #253** at both sites — future readers see *why* the guard is intentionally absent and won't re-add it.

### ♻️ Reuse & Duplication (0)
- No duplication or over-engineering found. Tests reuse the existing `makeRunner` / `makeManager` / `makeApp` harness; the change is net-negative code (removes 2 validation blocks) plus tests. The two near-identical `#253` comments across router/manager are intentional per-file documentation, not code duplication.

### 🔴 Critical (0)
(none)

### 🟡 Medium (0)
(none)

### 🟢 Minor (1)
1. `tests/unit/cron-manager.test.ts` (T11b) — the "no delivery" assertion only checks that Telegram `/sendMessage` wasn't called. Since `makeManager` sets a `botToken` but the job has no Discord either, a symmetrical assertion that no Discord webhook fired would fully lock the "delivery is skipped" claim. Optional — the current assertion already fails if the fix regresses.

### ✅ AC Coverage (#253)
- [x] Channel-less `type=agent` create returns `201` (was `400`) — covered by `T23b`
- [x] Channel-less agent job runs and persists to Run History — covered by `T11b`
- [x] `prompt` still required for `type=agent` — covered by `T11e`
- [x] Delivery unchanged when a channel *is* present — existing `T23c` / delivery tests untouched
- [x] Create + update consistent (fix at `validatePayload` covers both) — covered by `T11c`/`T11d`

### 🧪 Gate (re-verified independently on the pushed tree)
- `tsc --noEmit` ✓
- `cron-router` + `cron-manager`: **62/62** passed

Small, correctly-scoped fix that aligns create-time validation with what the run engine already supports, with honest tests (positive + negative + regression guard). The single Minor is optional.

PR #279, opened 2026-08-08
feat: image generation (generate_image MCP tool) + media pipeline

0xthoth:
## Summary

Adds first-party image generation to claude-gateway via a `generate_image` MCP tool, plus the supporting media pipeline: a session-scoped media store, reference-image support (edit/continue an earlier result), a signed public media route, and a per-gateway `/shared` image route.

This branch continues the image-gen work that was first introduced upstream and layers on the following, all generic and env-driven (no deployment-specific hostnames or product names):

## What's included

- **`generate_image` MCP tool** — text-to-image and image-to-image, model list resolved from the configured provider.
- **Media store** — session-scoped artifacts persisted in the existing SQLite (`image_artifacts`), self-migrating on start; no manual migration.
- **Reference images** — resolve prior results by index ("Image N") / catalog metadata so a follow-up prompt can edit a previous image.
- **Signed public media route** — short-lived signed `/public/:token` URLs for external delivery (e.g. messaging webhooks), TTL via `GATEWAY_MEDIA_URL_TTL_MS` (optional).
- **`/shared` image route** — public-by-capability-token share links.
- **Security hardening** — redirect/SSRF screen on reference URLs, reference-count budget, absolute deadline on generation.

## Configuration (all optional / env-driven)

| Env | Purpose |
| --- | --- |
| `IMAGE_BASE_URL` / `IMAGE_API_KEY` | image provider endpoint + key |
| `GATEWAY_API_URL` | gateway self-reference for minting media URLs |
| `GATEWAY_MEDIA_URL_TTL_MS` | signed-URL lifetime (has a sane default) |
| `IMAGE_SHARE_MAX_REFS` | max reference images per request |
| `GATEWAY_SESSION_MEDIA_DIR` | on-disk media dir where generated images land — **auto-provisioned per API session by the gateway (pre-existing infra from #119); not an operator knob, listed for completeness only** |

## Notes

- `gateway.publicUrl` docs and all test fixtures use `example.com` placeholders — no real hostnames.
- No new required env: a gateway with none of the above set keeps working; the image tool simply reports no provider configured.
- SQLite `image_artifacts` self-alters on start — no manual migration step.


0xMaxMa:
### Follow-up: consolidated LINE `/public` into the single `/shared` bridge (`373bf7c`)

Heads-up on a larger change I pushed as a maintainer edit — happy to revert or split if you'd prefer.

**Why:** this PR shipped *two* parallel public-file systems: the stateful `/shared/:token` share bridge (image tools) and a separate stateless HMAC-signed `/public/:token` route (LINE image delivery). They duplicated the security-sensitive surface (path containment, magic-byte, XSS) and the `/public` HMAC path is what carried the token-forgery and provider-URL SSRF findings from review. Since `/public` is new in this PR (never released), now is the moment to collapse them into one before it ships.

**What changed:**
- **LINE now mints through `/api/v1/shares`** (the same bridge the image tools use) and builds `<base>/shared/<token>` from the webhook-derived `.public-base`. No more local HMAC signing.
- **`/public/:token` deleted** — route, `resolvePublicTokenKey`, `verifyPublicToken`/`signPublicToken`, `public-token.ts` (both copies), `public-media-headers.ts`. The forgery + SSRF surface goes with it.
- **`gateway.publicUrl` decoupled** from the store: it was doing double duty (URL base **and** on/off switch). Now the store + public route mount unconditionally; the private mint mounts when API keys exist; `publicUrl` is only the default base for the mint response `url`. Mint returns a **host-agnostic `token`**; `url` is an optional convenience omitted when `publicUrl` is unset — so a LINE-only pod with no `publicUrl` still works.
- **Neutral rename** to reflect it's a general primitive: `ImageShareStore`→`ShareStore`, `image-share-router`→`share-router`, `/api/v1/image-shares`→`/api/v1/shares`, `share-client` moved to `mcp/tools/shared/`.
- Kept the `.public-base` host derivation + the X-Forwarded-Host hardening from review.

Docs (API.md/README) updated. `tsc` clean; full unit suite 2633/2633 with new regressions for the `token` field, mint-without-`publicUrl`, and LINE-via-bridge.

Issue #363, opened 2026-08-20
Normalize HTML line-break tags in channel replies


0xMaxMa:
## Summary

Channel replies sent with `format: "html"` can render a literal `<br>` tag instead of a line break.

## Reproduction

Call the Telegram reply tool with:

```json
{
  "format": "html",
  "text": "first&lt;br&gt;second"
}
```

On current `main` (`016b2b4`), `containsTelegramHtml()` does not recognize `&lt;br&gt;`; the reply path then runs `toTelegramHtml()` while retaining HTML parse mode. The converter escapes the tag a second time and Telegram renders a literal `&lt;br&gt;`.

The same unsafe routing exists in both the direct tool path and receiver-server path. The automatic channel-forward path can show the same visible issue when assistant Markdown includes an escaped line-break tag.

## Proposed approach

Normalize `&lt;br&gt;` and `&lt;br/&gt;` to newline characters before format detection and conversion. Apply the normalization consistently to direct replies, receiver replies, automatic forwarding, and plain-text fallback conversion.

## Acceptance criteria

- `format: "html"` with `first&lt;br&gt;second` sends a visible line break, not a literal tag.
- Markdown mixed with an escaped break converts to valid Telegram HTML with a newline.
- Fallback-to-plain-text preserves the intended line break.
- Automatic forwarding does not expose escaped line-break tags.
- Regression tests fail on the pre-fix behavior.
- Explicit `format: "text"` remains unformatted.

PR #366, opened 2026-08-20
fix(telegram): preserve agent-authored Telegram HTML tags instead of escaping them

### LINKED ISSUE #365: Agent-authored Telegram HTML tags are escaped to literal text when output also contains markdown
## Summary
When an agent's final output contains **valid Telegram HTML tags written directly** (e.g. `<b>`, `<code>`, `<i>`, `<pre>`) *and* any markdown, the HTML tags are escaped to literal text and render as visible `<b>…</b>` in Telegram, while markdown-derived formatting (bold, links) still renders correctly. This is distinct from #363, which only normalized `<br>`.

## Symptom
A Telegram message shows literal `<b>`, `<code>` etc. as text, yet inline links from markdown `[text](url)` render as real hyperlinks in the same message — proving `parse_mode=HTML` is active but the agent's own tags were escaped before sending.

## Investigation (proven on v1.7.5)
Root cause is in `toTelegramHtml()` (`src/telegram/markdown.ts`, duplicated verbatim in `mcp/tools/telegram/lib/markdown.ts`). It is a hand-written tokenizer that recognizes markdown tokens and, for everything else, accumulates plain text and escapes it:

- `src/telegram/markdown.ts:241` — "Accumulate plain text until next markdown token"
- `src/telegram/markdown.ts:257` — `out.push(escapeHtml(text.slice(i, j)))`

An agent-authored `<b>` is not a markdown token, so it falls into the plain-text branch and `escapeHtml()` turns it into `&lt;b&gt;`. With `parse_mode=HTML` this displays as literal `<b>`.

The **reply** path already guards against this: `mcp/tools/telegram/module.ts` / `receiver-server.ts` compute `inputIsHtml = format === 'html' && containsTelegramHtml(text)` and send the text as-is when true. The **auto-forward** path has no equivalent — `src/agent/runner.ts:1780` calls `toTelegramHtml(channelText)` whenever `hasMarkdown(channelText)` is true, escaping any agent HTML.

### Live repro
```
node -e '
const { toTelegramHtml } = require("./dist/telegram/markdown.js");
console.log(toTelegramHtml("**bold** <b>x</b> [l](https://e.com)"));
'
```
Output (sent with `parse_mode=HTML`):
```
<b>bold</b> &lt;b&gt;x&lt;/b&gt; <a href="https://e.com">l</a>
```
`**bold**` → real `<b>`, markdown link → real `<a>`, but the agent's literal `<b>x</b>` is escaped to `&lt;b&gt;x&lt;/b&gt;` (renders as literal text).

## Scope / blast radius
- Deterministic: any agent final output (auto-forwarded) or reply that mixes markdown with directly-written Telegram HTML tags.
- Affects the auto-forward path and any `toTelegramHtml()` caller; both `src` and `mcp` copies are affected.
- Went unnoticed because agents that emit pure markdown never hit it, and pure-HTML replies (`format: html`) hit the `inputIsHtml` guard instead.

## Proposed approach
Make `toTelegramHtml()` **pass through valid Telegram-whitelist tags verbatim** instead of escaping them, while continuing to escape all other `<`/`>`/`&`:
- Whitelist: `b, strong, i, em, u, ins, s, strike, del, code, pre, blockquote, a, span, tg-spoiler` (open tags with optional attributes, and their closing tags).
- Add a tag-match branch to both the main token loop and the plain-text-accumulation break condition, in `src/telegram/markdown.ts` and the identical `mcp/tools/telegram/lib/markdown.ts`.
- Non-whitelisted tags (e.g. `<script>`) continue to be escaped as plain text — this is a controlled pass-through, not raw HTML injection.

### Tradeoff to note
After the fix, prose that literally discusses a tag name outside code (e.g. writing `<b>` as text, not inside backticks) would be interpreted as formatting. Text inside inline/blocks code is already escaped and remains safe.

## Acceptance criteria
- [ ] Mixed input `**bold** <b>x</b> [l](url)` renders bold, real hyperlink, **and** a real bold `x` (no literal `<b>`), verified via a unit test on `toTelegramHtml()`.
- [ ] Agent-authored `<code>`, `<i>`, `<pre>`, `<a href>` are preserved (not escaped).
- [ ] Non-whitelisted tags (`<script>`, `<div>`) remain escaped.
- [ ] Fix applied to both `src/telegram/markdown.ts` and `mcp/tools/telegram/lib/markdown.ts`; the auto-forward path (`src/agent/runner.ts`) renders agent HTML correctly.
- [ ] Regression test proven to fail on the pre-fix code.
- [ ] `npm run typecheck` and full `npm test` green.

0xMaxMa:
## Summary
Agent output that mixes markdown with **directly-written Telegram HTML tags** (`<b>`, `<code>`, `<i>`, `<a>`, …) rendered the HTML tags as literal `<b>` text, while markdown-derived formatting/links rendered correctly. Root cause and fix are distinct from #363 (which only normalized `<br>`).

## Root cause (proven on v1.7.5)
`toTelegramHtml()` (`src/telegram/markdown.ts`, duplicated in `mcp/tools/telegram/lib/markdown.ts`) is a hand-written tokenizer. Any non-markdown run is accumulated as plain text and `escapeHtml()`-escaped (`markdown.ts:257`), so an agent-authored `<b>` becomes `&lt;b&gt;` and, under `parse_mode=HTML`, shows as literal `<b>`.

Two paths were affected:
- **Auto-forward** (`src/agent/runner.ts`) only routed to `toTelegramHtml()` when `hasMarkdown()` was true. Pure-HTML output (no markdown) was sent as plain text → literal tags. Mixed output was escaped → literal tags.
- The **reply** path already had an `inputIsHtml` guard, but via a divergent inlined regex.

### Live repro (pre-fix)
```
node -e 'const {toTelegramHtml}=require("./dist/telegram/markdown.js");
console.log(toTelegramHtml("**bold** <b>x</b> [l](https://e.com)"))'
# => <b>bold</b> &lt;b&gt;x&lt;/b&gt; <a href="https://e.com">l</a>
```

## Fix
- **Preserve whitelist tags**: `toTelegramHtml()` passes valid Telegram tags (`a, b, blockquote, code, del, em, i, ins, pre, s, span, strike, strong, tg-spoiler, u`) through verbatim and escapes only the `<`/`>` that are *not* part of such a tag. Non-whitelist tags (`<script>`, `<div>`) are still escaped — a controlled pass-through, not raw HTML injection.
- **Auto-forward** now routes to `toTelegramHtml()` when the text has markdown **or** `containsTelegramHtml()` — so pure-HTML agent output renders too.
- **Consolidation**: `module.ts` and `receiver-server.ts` previously each carried their own inlined tag regex that listed a non-existent `<spoiler>` tag and omitted `<span>`. Both now use the shared `containsTelegramHtml()`. Applied to both `markdown.ts` copies.

## Tests
- New unit tests for `containsTelegramHtml()` and tag preservation in `toTelegramHtml()`, **proven to fail on the pre-fix code** (they produce the escaped `&lt;b&gt;` output).
- Covers: mixed markdown+HTML renders both, lone `<code>` with no markdown, `<b>/<i>/<a>` together, non-whitelist tags stay escaped, a tag inside inline code stays literal, and `a < c > d` stays escaped.
- `npm run typecheck` clean; `mcp` `tsc --noEmit` clean; full `npm test` green: **182 suites, 3448 tests, exit 0**.

## Tradeoff
After this change, prose that literally writes a tag name **outside** code (e.g. `<b>` in plain text) is interpreted as formatting. Tags inside inline/`code` blocks are still escaped and remain safe.

Closes #365

### DIFF
--- a/mcp/tools/telegram/lib/markdown.ts
+++ b/mcp/tools/telegram/lib/markdown.ts
@@ -294,3 +294,28 @@ export function toTelegramHtml(text: string): string {
 
   return out.join('')
 }
+
+/**
+ * Resolve how a channel reply should be sent given an optional explicit format.
+ * Mirrors the auto-forward decision in src/agent/runner.ts: agent-authored
+ * Telegram HTML tags must trigger HTML mode even when the text contains no
+ * markdown, otherwise a pure-HTML reply falls through to raw text with no
+ * parse_mode and the tags render literally (e.g. a visible "<b>").
+ * - explicit 'html' + already-valid Telegram HTML -> send verbatim under HTML
+ * - explicit 'html' (not yet valid) -> convert via toTelegramHtml under HTML
+ * - explicit 'text' -> always raw, no parse_mode (caller asked for plain text)
+ * - auto (no format) -> HTML when markdown OR Telegram tags are present
+ */
+export function resolveTelegramReplyFormat(
+  text: string,
+  explicitFormat?: string,
+): { sendText: string; parseMode: 'HTML' | undefined } {
+  const inputIsHtml = explicitFormat === 'html' && containsTelegramHtml(text)
+  const useHtml =
+    inputIsHtml ||
+    explicitFormat === 'html' ||
+    (!explicitFormat && (hasMarkdown(text) || containsTelegramHtml(text)))
+  const sendText = useHtml && !inputIsHtml ? toTelegramHtml(text) : text
+  const parseMode: 'HTML' | undefined = useHtml ? 'HTML' : undefined
+  return { sendText, parseMode }
+}
--- a/mcp/tools/telegram/lib/markdown.ts
+++ b/mcp/tools/telegram/lib/markdown.ts
@@ -8,6 +8,27 @@ export function normalizeTelegramLineBreaks(text: string): string {
   return text.replace(/(?:<|&lt;)br\s*\/?(?:>|&gt;)/gi, '\n')
 }
 
+// Opening (with optional attributes) or closing form of every tag Telegram's
+// HTML parse mode honours. Sticky so it can be tested at a given offset without
+// slicing. `[^<>]*` keeps an attribute run from swallowing past the tag.
+const TELEGRAM_HTML_TAG = /<\/?(?:a|b|blockquote|code|del|em|i|ins|pre|s|span|strike|strong|tg-spoiler|u)(?:\s[^<>]*)?>/iy
+
+/** Length of a valid Telegram HTML tag starting at `i` (where text[i] === '<'), or 0 if none. */
+function matchTelegramTag(text: string, i: number): number {
+  if (text[i] !== '<') return 0
+  TELEGRAM_HTML_TAG.lastIndex = i
+  const m = TELEGRAM_HTML_TAG.exec(text)
+  return m ? m[0].length : 0
+}
+
+/** Whether text already contains at least one Telegram-whitelist HTML tag written directly. */
+export function containsTelegramHtml(text: string): boolean {
+  for (let i = text.indexOf('<'); i !== -1; i = text.indexOf('<', i + 1)) {
+    if (matchTelegramTag(text, i) > 0) return true
+  }
+  return false
+}
+
 /**
  * Detects whether text contains markdown formatting patterns
  * that warrant HTML rendering in Telegram.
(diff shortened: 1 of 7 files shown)


0xMaxMa:
Added the critical fix: the agent-facing reply tool is served by `mcp/tools/telegram/module.ts` (loaded by `mcp/server.ts` under bun), not `receiver-server.ts`. Its auto-mode format detection checked only `hasMarkdown(text)`, so a report authored with Telegram HTML tags but no markdown was sent raw with no `parse_mode` and rendered tags literally.

All three send paths now share one pure helper `resolveTelegramReplyFormat()` (module reply tool, receiver-server reply, runner auto-forward). Verified live: restarted MCP server runs the patched `module.ts`; the exact report content from the bug now resolves to `parseMode=HTML` with tags preserved.

Gates: tsc (src+mcp) clean, markdown 70/70 including regression proven-fail-on-old.

PR #367, opened 2026-08-21
feat(slack): add Slack channel

0xthoth:
## Summary

Adds Slack as a supported channel, mirroring LINE's shape: pure access
gate, HTTP webhook via the existing generic `/webhooks/:app/:agentId`
ingress, a `slack_reply` MCP tool, and DM/channel policy + pairing +
allowlist support consistent with the other channels. No reply-token TTL
to work around (unlike LINE) — `chat.postMessage` works with the bot
token at any time.

Built and tested end-to-end against a real Slack workspace (DM, channel
`@mention`, and thread replies all confirmed working), then run through
an independent code review that surfaced a further round of fixes —
see the second commit for details.

Rebased onto current `main` (this branch started diverging a while
back); verified with a full test-suite diff against a clean `main`
checkout — identical failure counts before/after (461 failing / same
suites, all pre-existing environment gaps unrelated to this change —
missing `fts5` sqlite support and sandboxed PTY timing tests in this
environment), with only new Slack tests adding to the pass count.

## What's in each commit

**`feat(slack): add Slack channel`** — types (`AgentConfig.slack`), pure
access gate (`slack-access.ts`, mirrors `line-access.ts`), webhook router
(HMAC signature verification, the POST-based `url_verification`
handshake, event normalization, 3-second ack), `SlackClient` wrapping
`chat.postMessage`/`reactions.add`/`reactions.remove`, `slack_reply` MCP
tool, wiring into the existing webhook dispatcher, unit tests.

**`fix(slack): correct source-type gaps`** — found live against a real
workspace: several `'telegram' | 'discord' | 'line'` value-computing
branches were widened at the type level when Slack was added but not at
the value level, so Slack sessions silently behaved as Telegram sessions
end to end (session store paths, reply-tool recognition, state
directories). Also fixed `app_mention` events carrying no `channel_type`
(unlike `message` events), and switched `SlackClient` to form-urlencoded
bodies uniformly (`users.info`/`conversations.info` don't reliably parse
JSON).

**`fix(slack): address code-review findings`** — an independent review
pass caught several more gaps: the plain-text/system-notice fallback
path had no delivery mechanism for Slack at all (silently dropped
`/stop`, `/rename`, connection-drop notices, etc.); a duplicate
`message_id` XML attribute bug affecting every channel; `thread_ts`
captured from inbound events but never surfaced to the agent, so
`slack_reply`'s `thread_id` could never actually be used; no
idempotency check on Slack's at-least-once event delivery; an
`isCliChannel()` runtime guard missing `'slack'` despite the type
already including it; and a previously undocumented/untested
`verifySlackSignature` (the HMAC check gating every inbound request).

**`refactor: consolidate the telegram/discord/line/slack union`** — the
literal `'telegram' | 'discord' | 'line' | 'slack'` union (and its
`| 'api'` variant) was hand-copied into roughly a dozen call sites
across 9 files. That exact duplication is what caused the type/value
mismatch bugs above — the type checker was satisfied while a
hand-maintained runtime branch quietly disagreed. Introduces one
canonical `CHAT_CHANNELS` const array in `history/types.ts`, with both
the `ChatChannel` type and an `isChatChannel()` runtime guard derived
from it, so a future channel addition updates every consumer's type
*and* guard together instead of relying on remembering to touch both.
Also collapses two other duplicated ternaries (a per-channel state
directory suffix repeated 3x, a directory-prefix scan hardcoding 5
channel names) into lookups derived from the same array.

## How to test

1. Configure `slack.botToken` + `slack.signingSecret` on an agent, point
   a Slack app's Event Subscriptions Request URL at
   `/webhooks/slack/<agentId>` (Socket Mode must be off).
2. DM the bot → pending-sender pairing flow → approve → confirm the
   reply arrives.
3. `@mention` the bot in an allowlisted channel → confirm it replies
   (needs the `app_mention`/`channel_type` fix).
4. Reply inside an existing Slack thread → confirm the bot's answer
   stays in-thread (needs the `thread_ts` fix).
5. `npm test` — full suite; diffed clean against `main` (see Summary).


0xMaxMa:
Code review follow-up — pushed `fa54b23` addressing the review findings:

- **Strip bot self-mention** from `app_mention` text in `normalizeSlackEvent` (bot id resolved from `authorizations[0].user_id`); only the bot's own `<@Ubot>` / `<@Ubot|label>` is removed, other users' mentions preserved.
- **Thread-aware auto-forward**: the Slack fallback path (command replies, socket-drop notices, no-tool-call assistant fallback) now replies in-thread when the last inbound message was threaded, mirroring `slack_reply`; a top-level message clears the tracked thread.
- **API.md**: added a `### Slack (app: "slack")` section parallel to LINE (setup, url_verification handshake, signature scheme, error table, access control, `/slack/pending` admin endpoints).
- **README**: updated the stale channel list.

Added `tests/unit/slack-normalize.test.ts` (mention stripping). Gates: `tsc` clean · slack suites 87/87 · agent-runner + channel 153/153.

Note: the `/stop` forcePersist change was reviewed and kept as-is — it's a correct improvement (web Stop button and typed `/stop` now leave the same visible outcome, no dangling user turn), not a defect.

PR #377, opened 2026-08-24
fix(session): stop /compact from freezing the whole gateway on large sessions

### LINKED ISSUE #376: [Bug] /compact freezes the entire gateway on large sessions (spawnSync + sequential chunks + no failure recovery)
## Summary
`/compact` on a large session (thousands of messages) freezes the **entire gateway** — every agent, every channel — for minutes, and can fail the whole job over a single transient model-call error. Root cause: `SessionCompactor.callClaudeForSummary` (`src/session/compactor.ts:160`, pre-fix) calls `spawnSync` to invoke `claude --print` once per ~100K-char chunk of history, **sequentially**, each with a 300s timeout. `spawnSync` blocks the daemon's single Node.js event loop for its full duration, and every agent in this gateway shares one process — so one compaction call freezes all of them until it returns.

## Investigation (live evidence)
Measured against a real production session (`deploy` session, agent `meguri`, session id `96a77d1f...`):
- **2,153 messages**, ~1.58M chars of history to summarize (~395K tokens)
- Chunked at `CHUNK_CHARS = 100_000` chars → **16 chunks**
- Old code: 16 sequential `spawnSync` chunk calls + 1 sequential merge call = **17 blocking model calls in a row**, each able to take up to the 300s (`SUBMIT`-style) timeout
- Worst case ≈ **15 minutes of total gateway freeze** for one `/compact` invocation — matching the observed symptom (`Could not connect to gateway`, then `Claude has not responded in 5 minutes` on unrelated sessions/channels)
- If **any single chunk call** fails or times out, `callClaudeForSummary` throws and the whole `/compact` job aborts with `Compact failed` — the user loses the entire ~15 minutes of work and must retry from scratch, with the same risk of hitting the same failure

The blocking mechanism is a known anti-pattern already documented elsewhere in this codebase — `src/agent/skill-learning/reviewer.ts:9` and `src/agent/dreaming/reviewer.ts:8` both call out: *"async spawn only … NEVER `spawnSync` (freezes the daemon event loop)."* Every other model-calling code path in the gateway already uses an async spawn seam (`makeClaudeSpawn`); `compactor.ts` was the sole remaining caller of `spawnSync`.

## Proposed Approach
Fix all four compounding defects in one PR (they're the same code path and the same root fix — chunked model calls done wrong):

1. **Async spawn** — replace `spawnSync` with an async `child_process.spawn`-based call (mirrors the existing `ClaudeSpawnFn` seam pattern in `reviewer.ts`), so the gateway's event loop is never blocked while a compaction chunk call is in flight. This is the fix for the actual freeze.
2. **Bounded-concurrency chunk summarization** — summarize chunks with a small worker pool (cap ~4 in flight, mirroring the existing pattern in `src/apps/installer.ts` `restoreRunningApps`) instead of one-at-a-time. Cuts wall time roughly by the concurrency factor on large sessions (~15min → ~3-4min for the measured case) without unbounded concurrent model calls.
3. **Per-chunk retry + graceful degrade** — retry a failed chunk once; if it still fails, fall back to a truncated raw excerpt for that chunk instead of aborting the entire compact job. Same treatment for the final merge-summaries call: retry once, then fall back to the concatenated (unmerged) per-chunk summaries rather than losing the whole compaction. A single bad model call can no longer waste the whole job.
4. **Progress visibility** — report chunk-summarization progress back to the user during a multi-chunk compact (throttled status updates, not per-chunk spam) so a large `/compact` doesn't look dead while it's still working.

## Acceptance Criteria
- [ ] `SessionCompactor` no longer calls `spawnSync`; the daemon event loop is not blocked during compaction (other agents/sessions remain responsive during a large `/compact`)
- [ ] Chunk summarization runs with bounded concurrency (not fully sequential, not unbounded)
- [ ] A single chunk's summarization failure (after retry) degrades to a raw excerpt instead of failing the whole compact job
- [ ] A merge-step failure (after retry) degrades to the concatenated per-chunk summaries instead of throwing
- [ ] The channel `/compact` command posts throttled progress updates during a multi-chunk compact
- [ ] Regression tests prove each behavior against the OLD code (proven-red): sequential-only concurrency, single-chunk-failure aborting the whole job, and merge-failure aborting the whole job
- [ ] All existing compactor/command/session test suites remain green; no behavior change for the non-chunked (small session) `/compact` path

## Out of Scope
- Auto-compacting a session automatically at a context-usage threshold (a separate, structural feature to prevent sessions from growing this large in the first place) — deferred to a future issue.

0xMaxMa:
## Summary
Fixes #376 — `/compact` on a large session freezes the entire gateway (all agents/channels) and can lose all progress to a single transient failure.

Root cause: `SessionCompactor.callClaudeForSummary` called `spawnSync` to invoke `claude --print` once per ~100K-char chunk of history, sequentially. `spawnSync` blocks the daemon's single Node.js event loop; every agent in this gateway shares one process, so one compaction call froze all of them until it returned. Measured against a real 2,153-message session: 16 chunks → 17 sequential blocking model calls → up to ~15 minutes of total gateway freeze per `/compact`, and any single chunk failure aborted the whole job.

## Changes (`src/session/compactor.ts`, `src/agent/runner.ts`)
1. **Async spawn** — replaced `spawnSync` with an async `child_process.spawn`-based call (`defaultSpawnClaude`), mirroring the existing `ClaudeSpawnFn` seam pattern already used in `agent/skill-learning/reviewer.ts`. The event loop is never blocked while a compaction chunk call is in flight. The spawn is injectable (`ClaudeCliSpawnFn`) via the `SessionCompactor` constructor, matching the codebase's existing testability convention — no real `claude` binary needed in tests.
2. **Bounded-concurrency chunk summarization** — chunks now summarize through a small worker pool (cap 4 in flight), mirroring the pattern already used in `apps/installer.ts`'s `restoreRunningApps`. Results are written by index so output order matches chunk order regardless of completion order. Cuts wall time roughly by the concurrency factor on large sessions.
3. **Per-chunk retry + graceful degrade** — a chunk gets one retry; if it still fails, it degrades to a truncated raw excerpt (clearly labeled) instead of throwing and killing the whole compact job. Same treatment for the final merge-summaries call: retry once, then fall back to the concatenated per-chunk summaries rather than losing the whole compaction.
4. **Progress reporting** — the channel `/compact` handler now posts throttled status updates (~25% steps) during a multi-chunk compact, so a large session doesn't look dead mid-compact. (Scoped as periodic status messages via the existing `writeAutoForward` path rather than true message-editing — the gateway has no per-channel message-ID/edit plumbing for this call site today, and adding one would expand this fix into channel-adapter territory across Telegram/Discord/Slack/LINE. Periodic status updates satisfy the same goal — visible progress, no "is this dead?" — without that scope creep.)

Scoping note: retry/degrade only applies to the **chunked** path (session large enough to need >1 chunk). The non-chunked (small session, single call) path is intentionally left throwing on failure as before — there's nothing to degrade against for a single call, and this keeps the existing small-session test contract unchanged.

## How to test
- `/compact` a small session (<5 messages) → still throws `NotEnoughMessagesError` as before.
- `/compact` a normal session (fits in one chunk) → unchanged behavior; a CLI failure still aborts with the original error message.
- `/compact` a very large session (>100K chars of history to summarize) → multiple chunk calls run concurrently (cap 4), progress messages appear every ~25%, and a single bad chunk/merge call no longer aborts the whole compaction.

## Regression tests (all proven-red against the pre-fix code before verifying green)
Added to `tests/unit/session-compactor.test.ts` under `large-session chunking (#376)`:
- `U-CMP-LG-2`: max concurrent chunk-summary calls is bounded by `CHUNK_CONCURRENCY` but > 1 (proves it's not sequential). **Proven red**: forcing `CHUNK_CONCURRENCY = 1` fails the `> 1` assertion.
- `U-CMP-LG-3`: a failing chunk degrades instead of failing the whole job. **Proven red**: removing the retry/degrade loop makes the whole compact throw.
- `U-CMP-LG-4`: a failing merge call degrades to the concatenated chunk summaries. **Proven red**: removing the merge retry/degrade makes the whole compact throw.
- `U-CMP-LG-5`/`U-CMP-LG-6`: `onProgress` fires once per completed chunk (and never for a non-chunked compact). **Proven red**: removing the `onProgress` call leaves the callback uninvoked.

Also rewrote the existing `spawnSync`-mocking tests to use the new injectable `ClaudeCliSpawnFn` (same behavior/assertions preserved, `U18`–`U21c`, `U-CB1/2`, `U-CMP-API-1..4` all still pass), and updated `tests/unit/api-session-commands.test.ts`'s `child_process` mock (its `spawn` mock now also resolves the compactor's `--print` calls, since compactor no longer uses `spawnSync`).

## Gates
- `tsc --noEmit`: clean
- `tsc` full build: clean
- Full suite: **3549/3549 passed**, 187 suites (`jest --runInBand`)
- Proven-red verified for all 4 fixes individually (see above), then restored + re-verified green

Closes #376

### DIFF
--- a/src/agent/runner.ts
+++ b/src/agent/runner.ts
@@ -2213,7 +2213,19 @@ export class AgentRunner extends EventEmitter {
 
     try {
       const compactor = new SessionCompactor(this.sessionStore);
-      const result = await compactor.compact(agentId, chatId, sessionId, compactModel, contextWindow, ch);
+      // Large sessions summarize in many chunks (see compactor.ts); post a status
+      // update at each ~25% step so the channel doesn't look dead mid-compact.
+      // Throttled (not per-chunk) to avoid spamming the chat on a 16+ chunk job.
+      let lastReportedPct = -1;
+      const onProgress = (done: number, total: number): void => {
+        if (total <= 1) return;
+        const pct = Math.floor((done / total) * 100);
+        if (pct - lastReportedPct >= 25) {
+          lastReportedPct = pct;
+          this.writeAutoForward(chatId, `⏳ Compacting session "${name}"... ${done}/${total} parts summarized`);
+        }
+      };
+      const result = await compactor.compact(agentId, chatId, sessionId, compactModel, contextWindow, ch, onProgress);
       await this.sessionStore.updateSessionMeta(agentId, chatId, sessionId, {
         loadedAtSpawn: undefined,
         archivedCount: undefined,
--- a/src/session/compactor.ts
+++ b/src/session/compactor.ts
@@ -236,18 +236,25 @@ export class SessionCompactor {
       }
     }
     const excerpt = chunk.length > DEGRADED_EXCERPT_CHARS ? `${chunk.slice(0, DEGRADED_EXCERPT_CHARS)}…(truncated)` : chunk;
+    console.error(
+      `[SessionCompactor] Chunk ${index + 1}/${total} summarization failed after ${CHUNK_MAX_ATTEMPTS} attempts (${lastErr?.message ?? 'unknown error'}); using truncated raw excerpt.`,
+    );
     return `[Part ${index + 1}/${total} — summarization failed after ${CHUNK_MAX_ATTEMPTS} attempts (${lastErr?.message ?? 'unknown error'}); showing truncated raw excerpt]\n${excerpt}`;
   }
 
   /** Merge per-chunk summaries into one; degrade to the unmerged concatenation on repeated failure. */
   private async mergeChunkSummaries(mergedText: string, model: string): Promise<string> {
+    let lastErr: Error | undefined;
     for (let attempt = 1; attempt <= CHUNK_MAX_ATTEMPTS; attempt++) {
       try {
         return await this.callClaudeForSummary(mergedText, model, MERGE_SUMMARIES_PROMPT);
-      } catch {
-        // retry, then degrade below
+      } catch (err) {
+        lastErr = err as Error;
       }
     }
+    console.error(
+      `[SessionCompactor] Merge-summaries call failed after ${CHUNK_MAX_ATTEMPTS} attempts (${lastErr?.message ?? 'unknown error'}); using unmerged per-chunk summaries.`,
+    );
     return mergedText;
   }
 
--- a/src/session/compactor.ts
+++ b/src/session/compactor.ts
@@ -1,6 +1,6 @@
 import * as fs from 'fs';
 import * as path from 'path';
-import { spawnSync } from 'child_process';
+import { spawn } from 'child_process';
 import { Message } from '../types';
 import type { ChatChannelOrApi } from '../history/types';
 import { SessionStore } from './store';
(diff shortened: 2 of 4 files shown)


0xMaxMa:
## Self-review

**Verdict: no blocking defects.** One real finding, fixed in-review.

**Checked and clean:**
- Worker-pool concurrency (`summarizeChunksConcurrently`): cursor-based work-stealing is race-free (JS single-threaded, cursor incremented synchronously before any `await`) — same pattern as `apps/installer.ts`'s `restoreRunningApps`. Results land by index, so output order matches chunk order regardless of completion order.
- `defaultSpawnClaude` timeout/close race: `settle()` is idempotent (`settled` flag), so a timeout firing `SIGKILL` and the subsequent `close` event racing can't double-resolve.
- `stdin` error handling: the async stream `'error'` event (e.g. EPIPE) is handled separately from the child's `'error'` event, same as the existing `makeClaudeSpawn` precedent in `reviewer.ts` — an unhandled stream error would otherwise crash the daemon.
- Env/PATH resolution (`pathWithNativeBin()`) preserved exactly as the old `spawnSync` call had it — no behavior change there.
- Retry/degrade scoping: correctly limited to the **chunked** path only. The non-chunked (small session) path still throws on failure as before — intentional, since there's nothing to degrade against for a single call, and it keeps the existing test contract for normal-sized sessions unchanged.
- Progress-callback throttling in `runner.ts` (`pct - lastReportedPct >= 25`) can't spam or overflow — `total` is always ≥2 whenever the callback fires (chunking only happens when `historyText.length > CHUNK_CHARS`, which `splitIntoChunks` always turns into ≥2 chunks).

**🟡 Found and fixed:** the per-chunk and merge-call graceful-degrade paths swallowed the failure with zero server-side signal — an operator debugging a low-quality compact summary would have no log evidence a degradation happened. Added `console.error` on both degrade paths (mirrors the existing precedent in `session/store.ts`'s corrupted-session-file recovery), plus regression assertions that the log fires.

**Docs:** No update needed — `/compact`'s external contract (endpoint, request/response shape, summary result) is unchanged; this is a purely internal implementation fix. Verified against `API.md`/`README.md`'s `/compact` references.

**Gates (after review-fix):**
- `tsc --noEmit`: clean
- `tsc` full build: clean
- Full suite: 3549/3549 passed pre-review-fix; all directly-affected suites (`session-compactor`, `api-session-commands`, `dreaming-compactor`, `api-command-persistence`, `api-stream-crash-recovery`) re-verified green post-review-fix (67/67)
- Proven-red re-verified for the new log-assertion tests as part of the fix cycle

PR #379, opened 2026-08-24
[Feature] Add memory_shared_write/memory_shared_delete MCP tools for on-demand shared-KB writes

### LINKED ISSUE #378: [Feature] Add on-demand memory_promote MCP tool for explicit shared-KB writes
## Summary
Add a new `memory_promote` MCP tool so an agent can explicitly push one note to the shared, cross-agent KB on demand — today the *only* write path is the nightly automatic "dreaming" promotion, which is delayed and score-gated.

## Investigation
- The shared vault is written today only via `makeSharedPromoter()` (`src/agent/knowledge/shared-promote.ts:18-30`), which builds a content-hashed filename (`${agentId}-${reason}-${hash8}.md`) and calls `writeSharedNote()` (`src/agent/knowledge/shared-writer.ts:41-62`, atomic temp-write + rename, slugified filename — no path traversal).
- `makeSharedPromoter` is wired from exactly two callers: the nightly auto-applier (`src/index.ts:359`) and the dashboard manual-accept endpoint (`src/api/gateway-router.ts:1173`) — both gated on `sharedCfg.enabled && sharedCfg.mode === 'auto'` (`shared-promote.ts:24`).
- The underlying dreaming pipeline (`src/agent/dreaming/index.ts:69,333-338`, `src/agent/dreaming/accept.ts:145-235`) only promotes a durable `add` proposal that already cleared `promotionThreshold`/`minRecallCount` scoring — an agent has no way to say "push this specific note now."
- The read side already exists and is unconditionally available whenever the shared KB is *enabled* (independent of `mode`): `mcp/tools/memory/module.ts:16,104-106` (`memory_search` corpus `"shared"`/`"all"`) resolves the vault via `sharedDbPathFromEnv()` (`mcp/tools/memory/archive-reader.ts:39-41`, env `GATEWAY_SHARED_KB_DIR`). That env var is forwarded per-session from `resolveSharedKbDir()` (`src/session/process.ts:517-523`), which itself only checks `cfg.enabled` — confirming the "enabled but not auto-mode" state is already a first-class, supported state for reads.
- There's a direct structural precedent for a *write*-capable, scope-aware MCP tool: `skill_create` (`mcp/tools/skills/module.ts:36-62`) already accepts `scope: "workspace" | "shared"` and resolves a shared target dir from an env var (`GATEWAY_SHARED_SKILLS_DIR`) — no extra sandboxing beyond "this tool only ever writes inside one fixed directory."
- Freshness: after writing a note, it must be reindexed into `kb.sqlite` to become searchable. The existing mechanism is `spawnArchiveReindex()` (`src/agent/knowledge/reindex-spawn.ts:42-71`) — a debounced (30s/workspace), detached, fire-and-forget spawn of the compiled `reindex-cli.js`. It's currently only called from the Node-side gateway (`src/agent/runner.ts:1486`), never from the Bun-side MCP process.
- Constraint: `mcp/` is a separate deployable package (own `package.json`/`node_modules`) that must never import from `src/` (enforced by the existing `mcp-no-src-imports` guard). `mcp/tools/memory/archive-reader.ts` already documents itself as "a deliberate ... mirror of the query half of `src/agent/knowledge/archive-db.ts`; keep the two in sync" — the same mirroring pattern applies to the write side.

## Proposed Approach
1. Add a `memory_promote` tool definition to `mcp/tools/memory/module.ts` (alongside `memory_search`/`memory_get`), input `{ content: string, reason: string }`.
2. Add `mcp/tools/memory/archive-writer.ts` — a Bun-side mirror of `writeSharedNote`/`sharedNoteFilename` (`src/agent/knowledge/shared-writer.ts`), reusing the same content-hash naming convention as `makeSharedPromoter` so a promoted note is indistinguishable from (and dedupes against) one the nightly path would have written.
3. Gate the tool on `GATEWAY_SHARED_KB_DIR` being set (shared KB `enabled`) — the same gate `memory_search` already uses for corpus `"shared"`. Deliberately do **not** require `mode: "auto"`: this is an explicit, agent-initiated write, not automatic promotion. Document that distinction in the tool description so it isn't mistaken for a mode-bypass bug later.
4. After a successful write, trigger a reindex so the note is searchable immediately — invoke the compiled `reindex-cli.js` the same way `spawnArchiveReindex()` does (`reindex-spawn.ts:58-67`), reusing its 30s debounce. May need one additional env var forwarded from `src/session/process.ts` alongside `GATEWAY_SHARED_KB_DIR` so the Bun tool has enough shared-config context to invoke the CLI correctly — to be finalized during implementation.

## Acceptance Criteria
- [ ] `memory_promote` is visible whenever the shared KB is `enabled` (mirrors `memory_search`'s shared-corpus gate) and returns a clear "shared KB is not enabled" error otherwise — never a silent no-op
- [ ] Given `{content, reason}`, the tool writes into the shared vault's notes dir using the same content-hash naming as the nightly promoter, so a repeat call with identical content does not create a duplicate file
- [ ] Works when `mode: "propose"` (explicit call ≠ automatic promotion) while the nightly auto-promoter remains disabled in that mode — no change to nightly gating
- [ ] After a successful write, the note is discoverable via `memory_search` (`corpus: "shared"` or `"all"`) without a manual reindex step or waiting for the nightly dreaming run
- [ ] Unit tests: atomic-write helper (temp+rename, filename slugification/collision safety) mirroring existing `shared-writer` test coverage; tool-handler tests for enabled/disabled, content-hash idempotency, and the reindex trigger call
- [ ] No regression in existing `mcp/tools/memory` and `mcp/tools/skills` test suites

## Out of Scope
- Any change to nightly dreaming auto-promotion gating (`promotionThreshold`, `minRecallCount`, `maxChangesPerRun`)
- The dashboard manual-accept flow (`src/agent/dreaming/accept.ts`) — unchanged
- Any new cross-agent permission/ACL model — access to a project's shared vault stays all-or-nothing, exactly as `skill_create scope:"shared"` already behaves

0xMaxMa:
## Summary
Adds two MCP tools — `memory_shared_write(content, reason, force?)` and `memory_shared_delete(reason)` — so an agent can push or remove a note in the shared, cross-agent KB immediately, instead of relying solely on the nightly automatic "dreaming" promotion (which is delayed and score-gated). Mirrors the existing `skill_create(force)` / `skill_delete` pattern: `force:false` (default) fails if a note for that `reason` already exists (create semantics); `force:true` overwrites it (update semantics).

## Related Issue
Closes #378

## Changes Made
- `mcp/tools/memory/archive-writer.ts` (new): Bun-side write mirror of `src/agent/knowledge/shared-writer.ts`. Note identity is a stable `<agentId>-<reason>` slug (not a content hash), so repeated calls with the same `reason` update the same note and deletes are automatically scoped to the calling agent's own notes. Also adds `triggerSharedReindex`, a fire-and-forget spawn of the compiled `reindex-cli.js` so a written/deleted note is searchable (or gone) within seconds instead of at the next natural reindex.
- `mcp/tools/memory/module.ts`: registers `memory_shared_write` and `memory_shared_delete`, gated the same way `memory_search` gates `corpus:"shared"` — a clear "shared KB is not enabled" error, never a silent no-op. Works regardless of `mode` (`propose` or `auto`) since this is an explicit, agent-initiated write, not automatic promotion. `GATEWAY_AGENT_ID` is required to scope ownership — if it's ever unset, both tools fail closed with an explicit error rather than falling back to a shared default identity that could let two misconfigured agents overwrite or delete each other's notes.
- `src/session/process.ts`: forwards `GATEWAY_NODE_EXEC_PATH` (the gateway's own Node `execPath`) to the MCP subprocess env. This fixes a real bug caught while writing the e2e test: the Bun MCP process's own `process.execPath` points at the `bun` binary, and spawning the compiled reindex CLI (needs `node:sqlite`) with that fails silently (`No such built-in module: node:sqlite`) since the child runs detached with `stdio:'ignore'`.
- `mcp/tools/memory/archive-writer.test.ts` (new): unit coverage for the slugify/naming/atomic-write/delete helpers, plus an end-to-end test proving write → reindex → searchable via the real compiled CLI.
- `mcp/tools/memory/module.test.ts` (new): unit coverage for the `memory_shared_write`/`memory_shared_delete` fail-closed gate when `GATEWAY_AGENT_ID` is unset.
- `tests/integration/pipeline-mcp.test.ts`: asserts `GATEWAY_NODE_EXEC_PATH` is forwarded correctly in the written `mcp-config.json`.
- `README.md` / `API.md`: document the new on-demand write/delete path alongside the existing Shared KB description.

## Testing
- `npm run typecheck`: pass
- `npm run build`: pass
- `npm test`: 187 suites / 3549 tests pass
- `cd mcp && bun test tools/memory/`: 21/21 pass (includes a real end-to-end write→reindex→search test against the compiled CLI, plus the fail-closed ownership gate)
- Proven-red: reverted each core fix individually (filename sanitization, content-hash → slug identity, `GATEWAY_NODE_EXEC_PATH` forwarding, `GATEWAY_AGENT_ID` fail-closed gate) and confirmed the corresponding test fails on the pre-fix code, then restored.


0xMaxMa:
## Self-review

Reviewed the diff for correctness, security, necessity, duplication, and consistency.

**Found and fixed (2 real issues):**
1. **Duplication** — `archive-writer.ts` re-implemented `sharedNoteFilename`/`writeSharedNote` (slugify + atomic temp-then-rename) even though that logic has no `node:sqlite` dependency, and `mcp/` already imports similar Node-side logic straight from `dist/` elsewhere (`slack/module.ts`'s `SlackClient`, `telegram/receiver-server.ts`'s `turn-trace`/`incident-store`). Switched to importing the real functions from `dist/agent/knowledge/shared-writer.js` instead of maintaining a parallel copy that could drift out of sync.
2. **Missing size cap (security/necessity)** — `memory_promote` had no content size limit, unlike its direct structural precedent `skill_create` (`MAX_SKILL_SIZE`, 100KB). Since the shared vault is cross-agent, an uncapped write lets one agent (a runaway loop, or a prompt injection) exhaust shared disk / bloat the shared FTS5 index for every agent sharing that project. Added the same 100KB cap, mirroring `skill_create`'s exact pattern.

**Verified safe (no change needed):**
- Traversal: `reason`/`agentId` are interpolated into the raw note name before `sharedNoteFilename` slugifies the *whole* string — same containment property already covered by the Node-side test suite (`knowledge-shared.test.ts`).
- `GATEWAY_NODE_EXEC_PATH` forwarding: proven necessary — writing the e2e reindex test surfaced a real bug (the Bun MCP process's own `process.execPath` is the `bun` binary, which cannot run the `node:sqlite`-based reindex CLI; it failed silently since the child is detached with `stdio:'ignore'`). Fixed by forwarding the gateway's own Node `execPath`, with a `node`-on-PATH fallback.
- Synchronous fs writes in the tool handler (via `writeSharedNote`) are consistent with existing `memory_search`/`memory_get` sync I/O in the same file, and bounded by the new 100KB cap — not the `spawnSync`-style multi-minute blocking issue from a prior fix in this repo (#377); this MCP server is also per-session, not shared across agents.

**Gates (re-verified after the fixes):**
- `npm run typecheck`: clean
- `npm test`: 3548/3549 (one unrelated flake in `phase-manual.test.ts`, confirmed passing 51/51 in isolation — a known pre-existing flake under full-suite parallel load, unrelated to files this PR touches)
- `cd mcp && bun test tools/memory/`: 15/15
- `mcp-no-src-imports.test.ts`: passes with the new `dist/agent/knowledge/shared-writer.js`/`types.js` imports

PR is `MERGEABLE`/`CLEAN` at `44c0bcf`.

Issue #380, opened 2026-08-24
[Bug] Telegram auto-forward: single-slot `.forward` file + chat-level `.replied` marker can silently drop turn text


0xMaxMa:
## Summary
The Telegram delivery pipeline's auto-forward safety net (`mcp/tools/telegram/typing.ts`) can silently lose a turn's text when back-to-back turns overlap the typing loop's 4s poll tick — no error, no incident, message just never arrives.

## Investigation
Root mechanism (`mcp/tools/telegram/typing.ts`):
- `createWorkingStateManager().start(chatId)` is idempotent — if a `WorkingState` already exists for the chat, a new inbound message does **not** reset `state.startedAt` (`typing.ts:535-536`).
- `SessionProcess.sendMessage()` (`src/session/process.ts:1159-1168`) unconditionally rewrites the bare typing signal file with a fresh timestamp on every injected turn, regardless of whether the receiver's typing loop is still alive.
- `stop(chatId)` (`typing.ts:454-528`) is only invoked from the interval tick when the bare signal file is observed missing (`typing.ts:599-604`) — a 4s-granularity check.
- The `telegram_reply` MCP tool writes the dedup marker `${chatId}.replied` **as soon as the reply tool is called**, not at turn end (`mcp/tools/telegram/module.ts:369`, `receiver-server.ts:789`) — per the code comment at `receiver-server.ts:785`, "reply does not mean done, agent may continue working."
- `AgentRunner.writeAutoForward()` (`src/agent/runner.ts:2513-2545`) persists auto-forward text with a plain `fs.writeFileSync` to a single-slot `${chatId}.forward` file — not a queue. It has 6+ call sites across `runner.ts` (e.g. `:460`, `:949`, `:1889`, `:1910`, `:1960`, `:2354` for the 32MB-recovery notice).

**Race — turn A replies, turn B's forward is silently dropped:**
1. Turn A starts → `typingManager.start()` creates `WorkingState{startedAt: Ta}`.
2. Turn A calls the reply tool → `.replied` is written immediately (turn still in progress).
3. Turn A's session ends → `AgentRunner.writeTypingDone()` removes the bare signal file.
4. Before the next 4s poll tick observes the missing signal file and runs `stop()` (which would delete `.replied`), the user sends a second message. The receiver calls `typingManager.start()` again — it's a no-op because `states.has(chatId)` is still true, so `state.startedAt` stays at `Ta`.
[...]

PR #382, opened 2026-08-25
fix(telegram): turn-scope auto-forward dedup and queue .forward writes

0xMaxMa:
## Summary
Fixes claude-gateway#380 — the Telegram auto-forward safety net could silently drop a turn's text when back-to-back turns overlapped the typing loop's 4s poll tick, with no error and no logged incident.

## Root cause (proven in #380, re-verified against current `main` before implementing)
Two independent bugs in `mcp/tools/telegram/typing.ts` / `src/agent/runner.ts`, both live on `main`:

1. **Chat-scoped `.replied` marker.** `createWorkingStateManager().start()` is idempotent — a second inbound message for an already-active chat is a no-op and does not reset `state.startedAt`. `SessionProcess.sendMessage()` unconditionally rewrites the bare typing signal file on every injected turn. Combined, this lets a turn A `.replied` marker (written when the reply tool is called) survive until `stop()` finally runs — coalesced with a later turn B — at which point `stop()`'s dedup check (`fsApi.existsSync(repliedPath)`) sees the leftover marker and deletes turn B's `.forward` text without ever sending it.
2. **Single-slot `.forward` file.** `AgentRunner.writeAutoForward()` wrote via a plain `fs.writeFileSync` overwrite. Two forward-worthy events for the same chat before drain (two command replies, or the same race window) silently clobber each other with no trace.

## Fix
- `.replied` and each queued `.forward` entry now carry a `turnId` — the live typing-signal timestamp read at write time by both writers (`telegram_reply` in `module.ts`/`receiver-server.ts`, and `writeAutoForward` in `runner.ts`). `stop()` and `drainOrphanForwards()` in `typing.ts` only treat a forward entry as already covered by the reply tool when its `turnId` matches the `.replied` marker's `turnId` — a marker from a different turn (or a legacy/orphan entry with no turnId) never suppresses delivery, favoring an extra message over a silently dropped one.
- `writeAutoForward()` now reads the existing `.forward` file (if any), appends the new entry, and writes the array back atomically (tmp + rename) instead of overwriting — so multiple forward-worthy events for the same chat all reach the user, in order.
- `stop()` and `drainOrphanForwards()` drain every entry in the queue, applying the turn-scoped dedup per entry.
- Discord's independent `.forward` poller (`mcp/tools/discord/module.ts`) has no `.replied` dedup and was never subject to this race — but it reads the same shared `.forward` file, so its parser is updated to handle the new array format (with graceful fallback to the legacy single-object and plain-text shapes) so it isn't broken by the format change.
- Backward-compatible parsing throughout: an in-flight `.forward`/`.replied` file from just before this deploy (plain text, or the pre-queue single-object shape) is still read correctly.

## Regression tests
Added to `tests/unit/telegram-plugin/typing.test.ts`:
- `U-TY-08b` — turn B's forward is delivered despite turn A's leftover `.replied` marker (different `turnId`). **Proven to fail on the pre-fix code** (reverted the turn-scoping, test failed with 0 calls; restored, test passes).
- `U-TY-08c` — two queued `.forward` entries both deliver in order, no clobber. **Proven to fail on pre-fix code** (received `[]` instead of both messages).
- `U-TY-DR-03b` — same coverage for the orphan-forward drain path (`drainOrphanForwards`). **Proven to fail on pre-fix code**.
- Existing `U-TY-08` / `U-TY-DR-03` (same-turn dedup) updated to assert on matching `turnId`s, preserving the legitimate same-turn dedup behavior.

`tests/unit/agent-runner.test.ts` and `tests/unit/session-commands.test.ts` updated for the new queue file shape (array of `{text, format, turnId}` instead of a single object).

## Acceptance criteria (from #380)
- [x] A `.replied` marker from turn N is never treated as satisfying dedup for turn N+1's `.forward`.
- [x] Two `writeAutoForward()` calls for the same chat before drain both reach the user (in order), never silently overwriting each other.
- [x] New regression tests fail on the pre-fix code and pass on the post-fix code.
- [x] `drainOrphanForwards()` updated consistently with `stop()`'s new turn-scoped dedup and queue-draining logic.
- [x] Full test suite green; no regression in the existing `.replied`/`.forward` delivery tests.

## Out of scope (per #380, verified)
- LINE/Slack (already special-cased away from the `.forward`/`.replied` mechanism in `writeAutoForward()`).
- Discord's dedup racing — Discord has no `.replied` check at all (`processForwardFiles()` just delivers and deletes), so it was never subject to this race; only its `.forward` parser needed a compatibility update for the new shared file format.
- The turn-trace watchdog (`src/agent/turn-trace.ts`) — read-only observer, only checks `.replied` existence (unaffected by the content-shape change).

## Test plan
- [x] `npm run typecheck` — clean
- [x] `npm run build` — clean
- [x] `npm test` — 3552/3552 passed (187 suites)
- [x] Proven-red: all 3 new regression tests fail on the pre-fix `typing.ts`, pass after restoring the fix

### DIFF
--- a/mcp/tools/discord/module.ts
+++ b/mcp/tools/discord/module.ts
@@ -586,29 +586,37 @@ export class DiscordModule implements ChannelModule {
       if (!file.endsWith('.forward')) continue;
       const filePath = path.join(typingDir, file);
       const channelId = file.slice(0, -'.forward'.length);
-      let text: string;
-      let parseMode: undefined | 'html';
+      // AgentRunner.writeAutoForward() appends to a queue rather than
+      // overwriting (claude-gateway#380), so this file is a JSON array of
+      // { text, format, turnId } entries — but tolerate the pre-queue
+      // single-object form and the original plain-text form too, since a
+      // file in flight during a rolling deploy can be in either shape.
+      let entries: Array<{ text?: unknown; format?: unknown }>;
       try {
         const raw = fs.readFileSync(filePath, 'utf8').trim();
         try {
-          const parsed = JSON.parse(raw) as { text: string; format: string };
-          text = parsed.text;
-          parseMode = parsed.format === 'html' ? 'html' : undefined;
+          const parsed: unknown = JSON.parse(raw);
+          entries = Array.isArray(parsed) ? parsed : [parsed as { text?: unknown; format?: unknown }];
         } catch {
-          text = raw;
+          entries = [{ text: raw }];
         }
         fs.rmSync(filePath, { force: true });
       } catch { continue; }
 
-      if (!text) continue;
-      try {
-        const channel = await this.client.channels.fetch(channelId);
-        if (parseMode === 'html') {
-          // Discord doesn't support HTML — strip tags and send plain text
-          text = text.replace(/<[^>]*>/g, '');
-        }
-        await sendMessage(channel, text, {});
-      } catch { /* non-fatal */ }
+      // Discord has no `.replied` dedup mechanism (unlike Telegram) — every
+      // queued entry is delivered, in order.
+      for (const entry of entries) {
+        let text = typeof entry.text === 'string' ? entry.text : '';
+        if (!text) continue;
+        try {
+          const channel = await this.client.channels.fetch(channelId);
+          if (entry.format === 'html') {
+            // Discord doesn't support HTML — strip tags and send plain text
+            text = text.replace(/<[^>]*>/g, '');
+          }
+          await sendMessage(channel, text, {});
+        } catch { /* non-fatal */ }
+      }
     }
   }
 
--- a/mcp/tools/telegram/module.ts
+++ b/mcp/tools/telegram/module.ts
@@ -363,10 +363,18 @@ export class TelegramModule implements ChannelModule {
       }
     }
 
-    // Write .replied marker
+    // Write .replied marker, tagged with the live typing-signal timestamp as
+    // this turn's id — the receiver (typing.ts) only treats a queued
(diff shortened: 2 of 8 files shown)


0xMaxMa:
## Self-review

Reviewed the diff for correctness/security/necessity/duplication/consistency and checked API.md/README.md staleness (no public API surface touched — `.forward`/`.replied` are internal IPC files, not documented anywhere).

**Found and fixed one real gap:** `stop()`'s auto-forward block originally read the queue, delivered it (which can take a while across multiple chunks/retries), and only then removed `.forward`. A `writeAutoForward()` append landing in that window would be silently discarded once `stop()` finished — the same class of bug this PR fixes, just narrowed to a smaller race window. `drainOrphanForwards()` already removed the file *before* delivering for exactly this reason (see its existing comment); `stop()` now matches that ordering. Added `U-TY-09b`, proven to fail against the pre-fix ordering (asserted `rmSync` happens before the first `sendMessage`).

**Checked and found no issue:**
- No new external attack surface — `.forward`/`.replied` are written only by this agent's own processes (AgentRunner, the `telegram_reply` MCP tool), never from untrusted input.
- Discord's `.forward` parser update is a compatibility-only change (array + legacy single-object + legacy plain-text) — no dedup behavior added there, matching the issue's scoping (Discord was never subject to the `.replied` race).
- Backward compatibility verified for an in-flight file from just before this deploy in either direction (plain text, pre-queue single-object, or the new array format all parse correctly).

**Gates (after the review-fix):** `tsc --noEmit` clean, `npm run build` clean, full suite `3553/3553` (187 suites, one test added — `U-TY-09b`).

Issue #386, opened 2026-08-25
[Bug] Dreaming/shared-KB: fix duplicate-note promotion + populate the existing wiki-link graph


0xMaxMa:
## Summary
Nightly dreaming's auto-promotion into the shared KB creates a new near-duplicate note every time a recurring fact gets reworded, instead of updating the existing one; separately, the wiki-link graph dashboard (`/knowledge/graph`) is fully built but has no real data feeding it because nothing ever writes `[[wikilink]]` references.

## Investigation

1. `src/agent/knowledge/shared-promote.ts:18-30` (`makeSharedPromoter`) names every promoted note `${agentId}-${reason}-${hash8}` (sha256 of content, first 8 hex chars) and writes it via `writeSharedNote()` directly. This is a deliberate, documented divergence from the freeform-name scheme `memory_shared_create`/`_update` use (see the doc comment at `mcp/tools/memory/archive-writer.ts:41-56`, which explicitly says the hashed scheme "exists so unrelated automatic promotions never collide" with manual ones) — not a bug in isolation, but it has an unwanted side effect: because the slug embeds a content hash, the same recurring fact reworded slightly by the dream reviewer each night hashes differently and is written as a brand-new file every time, never updating a prior promotion of the same fact. `memory_shared_delete`'s own tool description documents the fallout: "This cannot delete notes the nightly dreaming pipeline promoted (different naming scheme)" (`mcp/tools/memory/module.ts`, near line 122).
2. The promotion path also never calls `findSimilarSharedNotes` (`mcp/tools/memory/archive-reader.ts:223-228`, an FTS5 OR-match near-duplicate search) — the near-dup nudge that `memory_shared_create` already runs (`mcp/tools/memory/module.ts:220-235`) for manual writes is skipped entirely for automatic ones.
[...]

PR #399, opened 2026-08-26
fix: promote only real facts to the shared KB, and let the GC see every note

### LINKED ISSUE #398: [Bug] Shared-KB auto-promotion fills the vault with index pointers, merges unrelated notes, and its lifecycle GC is inert
## Summary

Nightly dreaming's shared-KB auto-promotion is running end-to-end without errors, but the vault it produces is largely unusable, and the lifecycle GC that is supposed to clean it up cannot see almost anything.

Measured on a live vault of 40 notes:

- **30 of 40 notes (75%) contain nothing but `MEMORY.md` index-pointer lines** — links to the promoting agent's own local `memory/*.md` files. All **32 markdown link targets are dangling** inside the shared vault (0 resolve).
- The near-duplicate matcher treats **81% of the vault (39 of 48 chunks) as "similar"** to a typical seed, so promotion almost never creates a new note — it merges into whatever BM25 ranks first, which is frequently an unrelated note.
- `kb_sources = 40` but `kb_entry_lifecycle = 2`, so the staleness GC can only ever consider 2 notes. The other 38 are effectively immortal.

These compound: (1) produces low-signal notes, whose shared boilerplate tokens make (2) match everything, and (4) means nothing is ever reclaimed.

Filed together because the fixes touch the same promotion/index path and a fix to any one alone leaves the observable failure in place.

### 1. Promotion does not look at what the content actually is

`src/agent/dreaming/accept.ts:199-204` (and the nightly twin at `src/agent/dreaming/index.ts:333-338`) gate promotion on:

```ts
opts.sharedPromote && x.proposal.op === 'add' && x.proposal.content && x.proposal.tier !== 'episodic'
```

The memory-tiering convention is that adding a durable memory also adds a one-line pointer to the `MEMORY.md` index. That pointer op is itself a durable `add` with content, so it satisfies every condition above and is promoted verbatim. The resulting shared note is a link to a file that only exists inside one agent's workspace.

Observed shape of such a note:

```
- [some_local_memory.md](some_local_memory.md) — one-line hook text
```

Link-resolution check over the live vault: 32 link targets, 32 missing. Of the 10 notes that do carry real content, 5 were written deliberately through `memory_shared_create`, not produced by auto-promotion.

### 2. "Similar" has no minimum bar

`src/agent/knowledge/archive-db.ts:564-578` builds the near-duplicate query by OR-ing the first 12 tokens of the seed:

```sql
WHERE kb_chunks_fts MATCH ?   -- "tok1" OR "tok2" OR ... OR "tok12"
ORDER BY bm25(kb_chunks_fts) LIMIT ?
```

`src/agent/knowledge/shared-promote.ts:89-99` then treats **any** hit as a near-duplicate:

```ts
const similar = findSimilarSharedNotes(sharedCfg, `${name} ${content}`, 3);
if (similar.length > 0) {
  const [primary, ...related] = similar;
  writeCapped(sharedCfg, primary.name, mergeIntoNote(existing, content, related.map(r => r.name)));
  return;
}
```

Replaying the exact query against the live DB with a real seed from the most recent nightly run:

```
tokens used: add | pointers | to | two | new | feedback | memories | to | MEMORY | md | index | <first content token>
top 1  bm25 -12.148  <note about a CI runner permission blocker>
top 2  bm25  -3.705  <note about an unrelated intermittent issue>
top 3  bm25  -3.098  <note about unrelated DB pool logic>
chunks matching this OR query: 39 of 48
```

The run merged the promoted fact into top-1 and attached `[[wikilink]]`s to top-2 and top-3 — three unrelated topics fused into one note. The BM25 gap (-12.1 vs -3.7) shows a usable signal exists; nothing consumes it.

Two further sharp edges in the same path:
- `.slice(0, 12)` means a long `reason` consumes the entire token budget, so the actual content never reaches the query.
- `mergeIntoNote` attaches `Related:` wikilinks to every runner-up with no bar at all, seeding the graph with false edges.

Note: the Bun-side twin (`mcp/tools/memory/archive-reader.ts:79-86`) uses the same OR construction, but there its result is shown to a human who must pass `confirm: true`. The Node path merges unattended, so it is the one that needs a bar.

### 3. `reason` is used as note identity but was never specified as a name

`src/agent/dreaming/reviewer.ts:48` asks the model for `"reason":"why"` — free-form justification. `src/agent/knowledge/shared-promote.ts:75` uses that string as the shared note's identity (`name = p.reason.trim()`).

Result: note names that are edit instructions rather than topics, e.g. `insert-after-cron-config-before-<section>-memory-section`. Such a name never matches a future occurrence of the same fact, so the exact-name dedup branch (`sharedNoteExists`) rarely fires and the run falls through to the unbounded similarity branch above.

The schema already has the right precedent: episodic proposals must supply `topic` matching `^[a-z0-9-]{1,64}$` (`reviewer.ts:63`). Durable proposals have no equivalent.

### 4. Lifecycle rows are never backfilled, so the GC is inert

`src/agent/knowledge/indexer.ts:172-177`:

```ts
const hash = sha256(content);
const existing = db.getSource(rel);
if (existing && existing.hash === hash) {
  result.filesSkipped++; // hash-guarded: unchanged, no re-chunk
  continue;
}
```

The `continue` returns before lines 190-195 compute entry blocks and before `archive-db.ts:317` writes `kb_entry_lifecycle`. Lifecycle rows therefore only ever appear for files whose content changed **after** the lifecycle feature shipped. Live state: `kb_sources = 40`, `kb_entry_lifecycle = 2`.

`shared-staleness.ts` selects candidates exclusively from `db.listLifecycle()`, so 38 of 40 notes can never be aged out, and retrieval-driven restore can never apply to them either.

Constraint for the fix: `firstSeen` is normally the file mtime (`indexer.ts:218`). Backfilling with `Date.now()` would reset every note's age to zero and delay any reclamation by a further `staleTtlDays`.

### 5. GC cadence is coupled to the LLM consolidation cadence

`runSharedStalenessGc` is only reachable from `SharedReflectionManager.reflectOnce` (`src/agent/knowledge/reflection.ts:273`), which is scheduled weekly (`REFLECTION_DEFAULTS`: Sunday 04:00 UTC).

The two halves have very different cost profiles:
- staleness GC — wall-clock driven, no model call, cheap, and already documented in-code as needing to run regardless of the revision watermark.
- cluster consolidation — one model call per cluster, capped by `maxClustersPerRun`.

Coupling them means retrieval-driven restore has up to 7 days of latency: a note that was retired and then retrieved stays missing from the active set for up to a week.

## Proposed Approach

1. **Only promote facts, not index pointers.** Add a single content check in the shared promoter (the choke point both call sites already funnel through) that skips promotion when the content carries no standalone fact — i.e. every non-empty line is a markdown index-pointer bullet. This fixes the source of the junk without touching either dreaming call site.

2. **Require a real similarity bar before an unattended merge.** Keep FTS as recall, then apply a deterministic, corpus-independent overlap gate (normalized token sets, stopwords removed) between the seed and each candidate before merging or emitting a `Related:` wikilink. Below the bar, create a new note instead of merging. Also stop letting a long name starve the content out of the seed token budget.

3. **Give durable proposals a real identity.** Accept an optional `topic` (`^[a-z0-9-]{1,64}$`) on durable `add` proposals, mirroring the episodic contract, and prefer it as the shared note name. When absent, fall back to the reason, but reject names that are edit instructions rather than topics (skip promotion rather than create an unusable note).

4. **Backfill lifecycle rows for hash-skipped sources**, seeding `first_seen` from the stored source mtime so ages are preserved rather than reset.

5. **Split the reflection cadence.** Run the timer daily: staleness GC every run; LLM consolidation only on the configured day. Keeps the model spend at today's level while cutting retrieval-restore latency from up to 7 days to at most 1.

6. **One-off cleanup of the existing vault** is expected to follow from (1) + (4): once lifecycle rows exist for every note, the pointer-only notes age out through the normal GC path rather than needing a bespoke deletion script.

## Acceptance Criteria

- [ ] Content consisting solely of `MEMORY.md` index-pointer bullets is not promoted to the shared vault; content carrying a real fact still is.
- [ ] A promotion whose seed shares only generic tokens with an existing note creates a new note instead of merging into it.
- [ ] `Related:` wikilinks are only emitted for candidates that clear the same similarity bar.
- [ ] A long `reason` no longer prevents the promoted content from contributing to the similarity seed.
- [ ] A durable proposal carrying a valid `topic` is stored under that name; an instruction-shaped name with no `topic` is not promoted.
- [ ] After an index pass, every shared note has a `kb_entry_lifecycle` row, including notes whose content did not change.
- [ ] Backfilled rows preserve the original age (`first_seen` from source mtime), not the backfill time.
- [ ] Staleness GC runs daily; LLM consolidation still runs on its configured day only, and model spend per week does not increase.
- [ ] Existing broken state recovers without manual intervention (already-promoted pointer-only notes become GC-eligible).
- [ ] Regression tests for each of the above, each proven to fail against the pre-fix code.

0xMaxMa:
Closes #398

Nightly shared-KB promotion was running cleanly and producing a vault that was mostly unusable. Measured on a live 40-note vault: **30 notes (75%) contained nothing but `MEMORY.md` index pointers**, all **32 markdown link targets were dangling**, and `kb_entry_lifecycle` held **2 rows for 40 indexed notes**, so the staleness GC could never reclaim any of it.

## What was wrong

**Promotion never looked at what the content was.** The gate was `op === 'add' && content && tier !== 'episodic'`. The memory-tiering convention writes a fact to its own file *and* a one-line pointer into the `MEMORY.md` index — the pointer op satisfies every condition, so it was promoted verbatim, links and all.

**"Similar" meant "shares at least one word."** `searchSimilar` ORs the first 12 seed tokens; `makeSharedPromoter` then treated `similar.length > 0` as a near-duplicate and merged. Replaying the exact query against the live DB with a real seed from the last nightly run returned **39 of 48 chunks**, and the top three hits were precisely the note that got merged plus the two that got attached as `Related:` wikilinks — three unrelated topics fused into one note. The BM25 gap (-12.1 vs -3.7) shows a usable signal existed; nothing consumed it.

**`reason` was used as the note's identity** but is prompted as free-form justification, producing names like `insert-after-cron-config-before-<section>-memory-section` — a name no later occurrence of the same fact can match, which pushes every promotion into the unbounded similarity branch above.

**Lifecycle rows were never backfilled.** `reindexInto` `continue`s on an unchanged content hash *before* entry blocks are computed, so rows only ever appeared for files edited after the feature shipped.

**GC cadence was coupled to LLM consolidation.** Both halves ran weekly, so a note retired and then retrieved stayed missing from the active set for up to seven days.

## What changed

| Fix | Behavior |
| --- | --- |
| Content filter | Content that is only index-pointer bullets is not promoted. A fact that merely *contains* a link still is. |
| Merge bar | FTS stays recall-only; an unattended merge now requires token containment ≥ 0.5 against the candidate's full body. Below the bar the fact gets its own note. |
| Wikilink bar | `Related:` links clear a **lower** bar (0.25) than merges — a link is additive, a merge destroys the distinction between two facts. They are attached whether the fact merges or lands as a new note. |
| Seed budget | The FTS term budget is split between name and content, so a long `reason` can no longer starve the promoted fact out of the query. |
| Note identity | Durable `add`s may carry an optional `topic` slug (mirroring the episodic contract), preferred as the note name. An instruction-shaped fallback name is replaced by a content-derived one rather than dropping the fact. |
| Retired twins | A `stale__` twin of the promoted name is folded back in and removed — on both the create and the update path — because retired notes stay searchable and the GC's restore refuses once the name is live again. |
| Lifecycle backfill | Hash-skipped sources get their lifecycle rows, seeded from **source mtime** so real ages survive instead of restarting the TTL clock. Guarded by a hash comparison in SQL, so the blocks are only parsed when something is actually missing. |
| Invalidation cap | `staleness.maxInvalidationsPerRun` (default 50, oldest-idle first) so widening the GC's visibility cannot retire a whole vault in one night. Restores are never capped. |
| Cadence | The reflection timer fires daily: staleness GC every run, LLM consolidation only on `dayOfWeek`. Weekly model spend unchanged; restore latency drops from ≤7 days to ≤1. |
| Observability | Every declined promotion logs a reason — including the note-size cap and unexpected write failures, which used to be silent. |

Containment was chosen over a BM25 cutoff deliberately: BM25 is corpus-relative, so a threshold tuned on one vault silently changes meaning as the vault grows. Note that containment is scored against a **token-capped seed**, so the 0.5 bar means "half of the fact's leading topic words are already in this note", not half of the entire fact — documented at `MIN_MERGE_CONTAINMENT`.

Existing broken state recovers on its own: once every note has a lifecycle row, the pointer-only notes age out through the normal GC path — at most `maxInvalidationsPerRun` per night — instead of needing a cleanup script.

## Review rounds

This PR was reviewed three times; rounds two and three found defects in the fixes themselves, not only in the original code.

**Round two (10 findings)** — most consequentially: `Related:` links were computed *inside* the merge branch, so a fact scoring between the link bar and the merge bar became a disconnected graph node; containment was scored against a single 500-char FTS chunk, which penalised exactly the long, much-merged notes that most needed to match; retired `stale__*` notes were valid merge targets; `decideStaleness` had no per-run ceiling; and the lifecycle backfill guard counted rows *by path*, which `replaceSource` never prunes.

**Round three (7 findings)** — the retired-twin fold covered only the create path; the note-size cap and the promoter's catch-all still declined silently, contradicting this PR's own docs; the scheduler's re-arm floor did not prevent the duplicate run it claimed to (flooring a near-zero delay only postpones re-serving the same slot — it now rolls forward a full day); and `mergeIntoNote` dropped related links when the fact was already recorded.

## De-fix proof

Every behavioural fix was reverted individually against the committed tests; each turned its own test red, and green again on restore.

```
de-fix[pointer-filter        ] -> Tests: 1 failed
de-fix[merge-bar             ] -> Tests: 1 failed
de-fix[seed-budget           ] -> Tests: 1 failed
de-fix[instruction-name      ] -> Tests: 1 failed
de-fix[topic-identity        ] -> Tests: 1 failed
de-fix[lifecycle-backfill    ] -> Tests: 3 failed
de-fix[daily-gc-only         ] -> Tests: 1 failed
de-fix[reviewer-topic        ] -> Tests: 1 failed
de-fix[new-note-related      ] -> Tests: 1 failed
de-fix[full-body-containment ] -> Tests: 1 failed
de-fix[stale-not-merge-target] -> Tests: 1 failed
de-fix[stale-twin-restore    ] -> Tests: 1 failed
de-fix[invalidation-cap      ] -> Tests: 1 failed
de-fix[backfill-hash-guard   ] -> Tests: 1 failed
de-fix[content-derived-name  ] -> Tests: 1 failed
de-fix[skip-logging          ] -> Tests: 1 failed
de-fix[twin-fold-on-update   ] -> Tests: 1 failed
de-fix[twin-kept-when-capped ] -> Tests: 1 failed
de-fix[size-cap-logging      ] -> Tests: 1 failed
de-fix[write-failed-logging  ] -> Tests: 1 failed
de-fix[merge-keeps-edges     ] -> Tests: 1 failed
de-fix[scheduler-roll-forward] -> Tests: 1 failed
```

The merge-bar test is built so FTS recall is guaranteed non-empty (the two facts deliberately repeat a few words) — otherwise it would pass for the wrong reason.

### What is *not* covered by a test

Stated rather than demonstrated, so it is not mistaken for proven:

- **Lazy block parsing in `backfillLifecycle`.** Passing a thunk instead of an array avoids work; external behaviour is identical, so there is nothing to assert beyond the existing "repeat passes do not bump the revision counter" test.
- **Three comment/naming-only changes** — the `_agentId` rename, `startReflecting`'s docblock, and the containment-seed documentation. No behaviour, no test.

## Gates

- `npm run typecheck` — clean
- `npm run build` — clean
- `npx jest --runInBand` — **191 suites / 3650 tests passed** (was 3616 on `main`; +34)

## Docs

`README.md` and `API.md` updated for fact-only promotion, the two-tier similarity bars, the `topic` slug, retired-twin folding on both write paths, lifecycle backfill, the invalidation cap, the daily-GC / weekly-consolidation split, and the declined-promotion logging. `API.md` also had a stale pre-#386 claim that promoted notes use "a content-hashed filename" — corrected.


0xMaxMa:
## Independent review round — all 10 findings addressed (`e6713aa`)

An independent `/code-review` pass on `993fe42` found ten issues, including two in the guards this PR had just added and one factual error in the PR description above. All ten are fixed.

| # | Finding | Fix |
| --- | --- | --- |
| 1 | `Related:` links were computed only inside the merge branch, so a fact in the 0.25–0.5 band became a disconnected graph node | Links computed once, attached to both the merge and new-note paths |
| 2 | Containment scored against the matched FTS chunk, not the note — strictest for the long, much-merged notes that most need to match | `filterNearDuplicates` takes a text resolver; the promoter passes the full note body |
| 3 | No `stale__` awareness: retired notes were valid merge targets, and a recurring retired name stranded the retired copy while two files shared one `entry_hash` | Retired notes excluded as targets; a recurrence folds the retired body back and deletes the twin |
| 4 | First run after the backfill could retire every already-expired entry at once (no per-run cap) | `staleness.maxInvalidationsPerRun` (default 50), oldest-idle first; restores never capped |
| 5 | The `993fe42` guard stopped the writes but not the work — blocks were parsed as an eager argument | Blocks are now a thunk, evaluated only when the guard says so |
| 6 | That guard counted rows *by path*, but `replaceSource` never deletes lifecycle rows, so stale rows from old versions cleared the bar | Compares the current blocks' `entry_hash` values via `kb_chunks`, falling back to parsing when the DB predates `entry_hash` |
| 7 | An instruction-shaped `reason` dropped the fact entirely — including genuine facts like "Update the deploy section after the API change" | Falls back to a content-derived name; abandoned only when nothing nameable remains |
| 8 | Both skip paths were bare `return`s, making a silent halt undiagnosable | Optional logger; every declined promotion logs its reason. Wired at both call sites |
| 9 | `searchSimilar` docblock still claimed a 12-term cap and a Bun/Node invariant that no longer held | Rewritten: it is recall, not precision, and why the Bun twin's smaller budget is fine |
| 10 | Consolidation weekday read from the fire instant (a fire slipping past local midnight skips a week); an early fire could run a slot twice | Weekday decided from the slot's target instant; re-arm delay floored |

### Correction to this PR's description

The **Deviation from the issue** section above argues that holding wikilinks to the merge bar "would have deleted the graph edges issue #386 added." That was wrong: below the merge bar the code produced **no** edges either way, because `related` was unreachable there. The lower bar is still the right design — a link is additive, a merge destroys the distinction between two facts — but it only actually does anything as of this commit.

### De-fix proof

Each fix reverted individually against the committed tests:

```
de-fix[new-note-related     ] -> Tests: 1 failed, 24 skipped, 25 total
de-fix[full-body-containment] -> Tests: 1 failed, 24 skipped, 25 total
de-fix[stale-not-merge-target] -> Tests: 1 failed, 24 skipped, 25 total
de-fix[stale-twin-restore   ] -> Tests: 1 failed, 24 skipped, 25 total
de-fix[invalidation-cap     ] -> Tests: 1 failed, 13 skipped, 14 total
de-fix[backfill-hash-guard  ] -> Tests: 1 failed, 13 skipped, 14 total
de-fix[content-derived-name ] -> Tests: 1 failed, 24 skipped, 25 total
de-fix[skip-logging         ] -> Tests: 1 failed, 24 skipped, 25 total
```

**Not proven by a test:** finding 5 (laziness) is a work-avoidance change with no observable behavioral difference, and finding 10's tests cover the `isConsolidationDay` / `msUntilNextDailyTime` helpers rather than the scheduler wiring that consumes them. Both are stated rather than demonstrated.

### Gates

`npm run typecheck` clean · `npm run build` clean · `npx jest --runInBand` **191 suites / 3643 tests** (was 3633; +10)

Docs updated for the per-run cap, content-derived naming, skip logging, full-body containment, and retired-note handling.

PR #424, opened 2026-08-31
fix(api): make streamed turns resumable across reconnects (#421)

### LINKED ISSUE #421: [Bug] Streamed turn cannot be re-attached: soft timeout and client disconnect both discard the rest of the turn
## Summary

A streamed turn on `POST /api/v1/agents/:agentId/messages` is bound to exactly one HTTP request. When the soft timeout fires, or the client disconnects, every remaining event of that turn is discarded and no client can ever get them back — even though the turn keeps running server-side and its result is persisted to history.

**Priority:** Medium

## Background / Context

Two client-visible symptoms, one root cause: **the gateway keeps no re-attachable state for an in-flight turn.**

### 1. The soft timeout is delivered as a terminal error

`src/agent/runner.ts:3546-3559` — when `opts.timeoutMs` elapses the gateway calls `notifyError(new Error('Agent response timeout'))`, which reaches the client as an SSE `{"type":"error"}` event. The turn itself is *not* cancelled: the output listener stays attached and the subprocess runs for up to a further `API_TIMEOUT_HARD_CAP_EXTRA_MS = 600_000` (`src/agent/runner.ts:81`) before `fail(new Error('Agent response timed out.'))`.

So the reply frequently *does* arrive and *is* persisted — but the client was already told the request failed and has no way to pick it up. Worse, the session stays locked for the whole soft→hard window: the same `session_id` answers `409` until the hard cap (documented at `API.md:1327`), so the client cannot even retry.

The comment at `src/agent/runner.ts:3541-3545` documents the intent — *"tell the SSE client now, but KEEP the output listener attached … so a turn that finishes a few seconds past the budget still lands in history"* — which is exactly right for history and exactly wrong for the client: history gains a message no live client will ever see.

### 2. A disconnected client can never re-attach

`src/agent/runner.ts:3461`:

```ts
const onClientDisconnect = () => { clientGone = true; };
```

Every emission is then guarded by `if (!clientGone)` (`src/agent/runner.ts:3477`, `3483`, `3494`, `3499`, `3503`, `3511`, `3517`). Nothing is buffered. There is no cursor, no replay, and no endpoint to attach a *new* SSE connection to an *existing* turn — the only way in is `POST …/messages`, which is rejected with `409` while the session is pending (`src/api/router.ts:476`, `runner.hasActiveApiSession`).

A browser reload mid-turn therefore loses the live stream permanently. The client's only recovery is to poll history until the turn lands — no partial text, no tool events, no progress.

### 3. The three SSE producers each hand-roll the same callbacks

`src/api/router.ts:428` (`/messages`), `:2577` (chat-session messages) and `:3275` (`/greeting`) all build an ad-hoc `sseCallbacks` object closing over `res`. Any buffering/replay must land in one place rather than being copy-pasted three times.

### Reported against a real client

The downstream web chat client requests `timeout_ms: 300_000` (600 s in image mode) and hits both symptoms: a turn that thinks past five minutes is reported to the user as a hard failure, and reloading the page mid-turn loses the stream for good. The client-side half is tracked in that client's own tracker.

## Acceptance Criteria

- [ ] While a turn is in flight, the gateway retains that turn's emitted `StreamEvent`s **in order**, each carrying a monotonically increasing sequence number, keyed by `session_id` + `request_id`
- [ ] A client can attach to an in-flight turn over a **new** SSE connection and receive, on that one connection: every event after a supplied cursor, followed by live events — with no gap and no duplicate at the seam
- [ ] Attaching with no cursor replays the turn from its first event
- [ ] The terminating `result` + `[DONE]` frames are delivered on the re-attached connection exactly as on the original one, including `attachments`
- [ ] A turn that **completed while no client was attached** is still retrievable through the same endpoint for a bounded grace window; after that the endpoint answers a distinct status (not `409`) telling the client to read history instead
- [ ] The soft timeout is no longer emitted as a terminal `{"type":"error"}`. It emits a distinguishable, **non-terminal** event (e.g. `{"type":"timeout","resumable":true,"request_id":…,"seq":…}`) so a client can re-attach rather than render a hard failure. `{"type":"error"}` remains reserved for genuine failures, including the hard-cap expiry
- [ ] Attaching to a turn is **not** treated as a conflict — `409` continues to mean "you tried to start a second turn", never "you tried to resume the first one"
- [ ] The retained buffer is bounded (per-turn cap) and freed when the turn ends plus its grace window; a long turn with heavy tool output cannot grow it without limit
- [ ] All three SSE producers (`src/api/router.ts:428`, `:2577`, `:3275`) share the buffering path — no fourth copy of the callbacks
- [ ] `API.md` documents the new endpoint, its cursor parameter, the new event type, and corrects the `409`/`504` notes at `API.md:1320-1327`
- [ ] Regression tests that fail on the current code: (a) a disconnect mid-turn followed by a re-attach yields the complete event sequence including `result`; (b) a soft timeout does not emit a terminal `error` and the turn is still resumable afterwards

## Out of Scope

- Any client-side change in a downstream web client — tracked in that client's own tracker
- Surviving a gateway process restart (the buffer is in-memory; a restart still loses the turn)
- Changing `DEFAULT_TIMEOUT_MS` (`src/api/router.ts:21`), the 600 s request cap (`src/api/router.ts:413`), or `API_TIMEOUT_HARD_CAP_EXTRA_MS`
- The non-streaming (`stream: false`) code path
- Telegram / Discord / LINE channels — this is the HTTP API surface only

## How to Implement

1. **Turn buffer.** In `AgentRunner` (`src/agent/runner.ts:214`, next to `pendingApiSessions`), add a per-session record holding `requestId`, a `StreamEvent[]` with sequence numbers, the terminal frame once known, and a completion timestamp. Populate it at the single point where events are produced, *before* the `clientGone` check — the guard should decide only whether to write to the socket, never whether to record.
2. **Stop dropping events on disconnect.** `src/agent/runner.ts:3461` should detach the socket sink, not silence the producer. Replace the seven `if (!clientGone)` call sites with a sink that is swappable at runtime, so a later re-attach can install a new one.
3. **Re-attach endpoint.** Add e.g. `GET /v1/agents/:agentId/sessions/:sessionId/stream?after_seq=N` next to the other session routes (`src/api/router.ts:3057-3220`). It should reuse the same SSE header block as `openSseStream` (`src/api/router.ts:447-455`), flush the buffered tail, then install itself as the live sink.
4. **Soft timeout.** At `src/agent/runner.ts:3546-3559`, replace the `notifyError(...)` call with the new non-terminal timeout event. Keep `errorNotified`/`settled` semantics for the hard cap, which remains a real error.
5. **Lifecycle.** Free the record where `pendingApiSessions.delete(sessionId)` already runs (`src/agent/runner.ts:3456`), after the grace window. Mirror the existing per-session attachment cleanup at `src/agent/runner.ts:3353` (`pendingApiAttachments.delete`) so a new turn never inherits a previous turn's buffer.
6. **Docs.** `API.md:1315-1327` (status table + notes) and the streaming section around `API.md:1654`.
7. **Tests.** Under `tests/`, alongside the existing API streaming coverage.

## Dependencies

- Blocked by: none
- Related to: the downstream web chat client's own issue for the client-side half (external tracker)

## Notes

- The two symptoms are one issue on purpose: a resumable turn buffer makes the soft timeout a non-event, and a soft timeout that is not terminal is only useful if there is something to resume onto. Splitting them would ship half a mechanism twice.
- Sequence numbers (not byte offsets) are the cursor because events are already discrete JSON frames; an `id:` line per SSE frame would additionally let a plain `EventSource` client use the standard `Last-Event-ID` header, which is worth considering.
- The hard cap should stay a real error: after 10 further minutes the turn genuinely is not coming back, and the client must be able to tell that apart from "still working".

0xMaxMa:
Closes #421

## The bug

A streamed turn was bound to exactly one HTTP request. Every event was written straight to `res` behind an `if (!clientGone)` guard, and **nothing was retained**, so the turn and the socket lived and died together:

1. **A disconnect mid-turn lost the rest of the turn permanently.** A browser reload, a flaky mobile network, a proxy idle timeout — `onClientDisconnect` set `clientGone` and every subsequent event was dropped on the floor. The turn kept running and its reply was persisted to history, but no client could ever see it stream.
2. **The soft timeout reported a failure for a turn that was still running.** It set the same flag and pushed a terminal `{"type":"error"}`, while the subprocess kept working for up to a further hard-cap window. The reply frequently *did* arrive and *was* persisted — to a client that had already rendered "request failed".

Both symptoms are the same missing thing: no re-attachable state for an in-flight turn.

## The fix

Decouple the producer from the socket.

`TurnStream` (new, `src/agent/turn-stream.ts`) records every event into an ordered, sequence-numbered buffer. The **sink** — the thing that writes to a socket — attaches, detaches and re-attaches independently. `attach()` replays from a cursor and installs the sink in **one synchronous block**, so no event can slip through the seam: no gap, no duplicate.

### New endpoint

```
GET /api/v1/agents/:agentId/sessions/:sessionId/stream?after_seq=N&request_id=…
```

Replays everything after the cursor, then keeps streaming live events on the same connection, terminating with the same `result` + `[DONE]` frames the original connection would have got.

- **Read access is enough** — resuming *reads* a turn, it does not start one.
- **Resuming is never a 409.** `409` still means "you tried to start a second turn on a busy session". A turn that cannot be resumed answers `410` with a discriminating `code`:

| `code` | Meaning |
|--------|---------|
| `TURN_GONE` | No turn for that session — never ran, or finished more than 2 min ago |
| `TURN_MISMATCH` | The session has a turn, but not the `request_id` you asked for |
| `TURN_TRUNCATED` | That cursor's events have been evicted |

### Other changes

- **Every SSE frame now carries `seq`** — the cursor a client passes back to resume.
- **The soft timeout is now a non-terminal event**: `{"type":"timeout","message":…,"resumable":true}`. The connection stays open and the turn keeps streaming. `error` stays reserved for genuine failures, including the hard-cap expiry.
- **Bounded buffers**: 2,000 events or ~4 MB per turn, whichever comes first, with amortised batch eviction. Completed turns stay replayable for a 2-minute grace window (which is what makes a reload right after the answer arrives still work). Release timers are `unref()`'d and cleared on `stop()`.
- **`API.md`** documents the endpoint, the `timeout` event, `seq`, the 410 codes, and corrects the now-stale `504`/`409` notes.

## Design decisions worth a second opinion

**A truncated cursor returns 410, not a partial replay.** Serving a replay with a silent hole in it would be worse than refusing — the client would have no way to know it was missing events. The cost is that a turn producing more than the cap is not resumable *from the start*; its recent tail still is, and the client can fall back to history.

**`id:` SSE lines / `Last-Event-ID` were deliberately skipped.** The issue lists this as "worth considering", not an acceptance criterion. Emitting `id:` lines would change the shape of every frame and could break naive clients that assume every frame starts with `data: `. The `seq` field carries the same information inside the payload, where existing parsers already look.

**`sendMessageToSession` (the channel/UI path) gets the resumable buffer but keeps its terminal timeout.** Its turns are now re-attachable, but its soft timeout semantics are untouched — that path isn't what #421 is about, and changing its failure contract belongs in its own change.

**A disconnected client's callbacks go quiet.** On disconnect the sink is detached, so the abandoned connection is not written to any more. Previously `onDone` still fired into the dead socket. Persistence is unaffected; the result is delivered to whoever re-attaches.

## Regression tests — and proof they fail on the old code

`tests/unit/api-stream-reattach.test.ts` (7), `tests/unit/turn-stream.test.ts` (12), `tests/unit/api-stream-endpoint.test.ts` (9).

I reverted the fix surgically — restored the "retain nothing" behaviour of the old `if (!clientGone)` guard, and restored the soft timeout's terminal `onError` — and re-ran. **8 tests failed, for exactly the right reasons:**

```
● a client that disconnects mid-turn re-attaches and gets the WHOLE rest of the turn

    - Array [ 2, 3 ]
    + Array []           ← nothing was buffered, so the replay is empty
```

```
● the soft timeout is a NON-terminal `timeout` event, not a terminal error

    expect(jest.fn()).not.toHaveBeenCalled()
    Received number of calls: 1
    1: [Error: Agent response timeout]
```

Restoring the fix returns all 28 to green.

One honest caveat: the disconnect/re-attach test necessarily calls `attachTurnStream`, which does not exist on the old code — so its *shape* is new. What the revert proves is that the underlying **behaviour** (retaining events past a disconnect) was absent, which is the thing the test is really asserting. The soft-timeout tests fail on the old code purely behaviourally, with no new API involved.

## Three pre-existing tests changed — please look at these

Two encoded contracts this change intentionally breaks, and I updated them rather than working around them:

- **`T13: timeout calls onError after timeout`** → `T13: soft timeout emits a non-terminal timeout event, not onError`. This is precisely the behaviour #421 asks to change. It now also asserts `onError` is *not* called and the session is still in flight.
- **`T14b: response persisted to history DB after client disconnect`** — its assertion that `onDone` fires after disconnect no longer holds (the sink is detached). Its actual subject, history persistence, is untouched and still asserted; I added an assertion that the result is retrievable via re-attach, which is where it goes now.
- **`stream: a result arriving AFTER the soft timeout still persists to history`** (#75) — waits on the `timeout` chunk instead of the error. Everything it actually checks is unchanged.

## Also fixed: a half-open SSE response

While running the full suite, 7 `api-greeting` tests hung. Root cause is in this PR's code but is a real bug, not test fallout: a throw while building the result frame was swallowed by the surrounding `try/catch`, leaving the response **unterminated** — any client waiting for `[DONE]` would hang until its own timeout. `res.end()` now runs in a `finally`.

## Verification

- `tsc --noEmit` — clean
- `jest` — **4073 passed, 228 suites, 0 failed**

---

# Follow-up: the AC deltas from the downstream-client investigation

[This comment](https://github.com/0xMaxMa/claude-gateway/issues/421#issuecomment-5474783588) added five acceptance criteria. Two were already satisfied by the work above; the third commit (`f4c542f`) closes the remaining three.

| AC | Status | Where |
|----|--------|-------|
| 1. A hard-capped turn does not leave a `⚠️` row *and* a reply row | ✅ `f4c542f` | `persistFailedApiTurn` |
| 2. The hard cap interrupts the turn instead of silently detaching | ✅ `f4c542f` | both hard-cap timers in `runner.ts` |
| 3. The hard cap is observable to a re-attached client | ✅ already, `78bba8c` | `errorNotified` was deleted outright |
| 4. The SSE `error` event carries `code` | ✅ `f4c542f` | `types.ts`, `turn-stream.ts`, `router.ts` |
| 5. The re-attach stream replays from the start of the turn | ✅ already, `78bba8c` | `attach(sink, afterSeq)`, default `0` |

### AC-1 + AC-2 are the same design question, answered by interrupting

The comment offers a choice: let the capped turn keep running and have its late result supersede the `⚠️` row, or genuinely stop it. This PR stops it, for the reason the comment itself gives — *"point 2 is the one that quietly wastes the most tokens"* — and because after `78bba8c` the soft→hard window is already a 10-minute grace period in which the turn stays fully resumable through `…/stream`. A turn still unfinished past that point is hung, not slow.

Interrupting also removes the contradiction AC-1 describes rather than papering over it: an interrupted turn produces no later reply, so history can never hold both rows for one turn. The `⚠️` row becomes simply true, and clients that use it to stop their spinner are right to.

**The trap worth reviewing.** `SessionProcess.interrupt()` gates on `_processing` and returns `false` without sending a signal once that flag is clear — and the existing code cleared it first:

```ts
session.setProcessing(false);
fail(...);              // ← adding session.interrupt() below this line does nothing
```

Reversing the two lines type-checks, reads correctly, and silently no-ops. `AC-2: interrupt() runs while the session still reports processing` asserts the **return value**, not just that the method was called, so the ordering cannot regress unnoticed.

### AC-1: the partial text stays

The capped row used to be `⚠️ Agent response timed out.` and nothing else, discarding whatever the turn had already streamed and the client had already rendered. It is now that text followed by the notice, in one row — the same shape `/stop` already produces.

### AC-4: `code` also survives a replay

`{ type: 'error'; message: string; code?: string }`, populated from the Error and written onto the SSE frame. The non-obvious half: a terminal frame reaches a *resumed* sink as buffered data, not as the live throw, so `callbackSink` was rebuilding a bare `new Error(message)` — exactly re-creating the string-matching problem for the clients most likely to need the code. The reconstructed Error now carries the frame's code.

Both API paths get the interrupt, streaming and sync, since they share the timeout chain.

**Deliberately left alone:** the crash path (`PROCESS_EXITED`) still persists only the notice, without partial text. Same defect shape as AC-1, different trigger from the one under review — happy to fold it in if you'd rather it were consistent.

### Verification for `f4c542f`

New file `tests/unit/api-hard-cap.test.ts` (5) drives a hung turn through both timeouts against a mocked subprocess; plus 2 in `turn-stream.test.ts` and 2 in `api-stream-endpoint.test.ts`.

Revert proof — reverting the three behaviours while keeping the tests fails **7**, for the right reasons:

```
● AC-1: the hard-cap row keeps the text the turn already streamed
    Expected substring: "Here is the first half of the answer"
    Received string:    "⚠️ Agent response timed out."
```

```
● AC-2: the hard cap SIGINTs the subprocess — a live turn is never silently abandoned
    expect(received).toContain("SIGINT")   ← no signal was sent at all
```

- `tsc --noEmit` — clean
- `jest` — **4086 passed, 229 suites, 0 failed**

---

## ⚠️ Breaking change — documented here, version left to you

The soft-timeout wire contract changed and needs a compatibility signal.

| | Up to and including 1.8.2 | With this PR |
|---|---|---|
| Turn passes `timeout_ms` | terminal `{"type":"error","message":"Agent response timeout"}`, stream ends | non-terminal `timeout` event, connection stays open up to the hard cap |

A client written against 1.8.2 or earlier that ignores unknown event types **stops failing at `timeout_ms`** and instead holds an open connection for up to 10 further minutes. `grep text_delta` finds no in-repo SSE consumer, so every affected client is external and cannot be migrated here — hence a Breaking callout at the head of the event-type table in API.md, with the migration step (handle `timeout`; close the connection yourself if you want the old cut-off).

**Recommendation, not applied:** this warrants a **minor** bump rather than a patch. `package.json` is untouched at `1.8.2` and the callout deliberately names no target version, so it stays correct whatever number the change ships under. `release.yml` is tag-triggered and publishes whatever `package.json` holds when `v*.*.*` lands, so whoever cuts the release owns the number.

## `ea95465` + `8e34bef` — remaining review findings

**Medium 2** — the break above: callout in API.md. (`ea95465` also bumped the version on its own; `8e34bef` reverts that and rewords the callout to be version-agnostic.)

**Minor 5** — `TurnStreamRegistry.start()` releases the session's previous record, so a completed turn's 2-minute replay window ends the moment a new turn starts. Two clients sharing a `session_id` (a phone resuming turn *N* while a laptop has already posted *N+1*) get `TURN_GONE` rather than the replay the grace window advertises. Now stated in the `TURN_GONE` row and beside the grace-window paragraph, with history as the documented fallback.

The added test pins the part that is **not** inherent to the map shape. Overwriting `turns.set(sessionId, …)` alone makes the old turn unreachable, so a test asserting only that passes even with `this.release(sessionId)` deleted. What `release()` actually buys is cancelling the completed turn's timer — which is keyed by session id, so a surviving one fires mid-flight and evicts the **live** turn:

```
● starting a new turn ends the previous turn's grace window immediately
    Expected: {… "requestId": "req-2" …}
    Received: undefined
```

Also splits a doc comment that `0075936` had joined onto `approxSize`'s signature line.

- `tsc --noEmit` — clean
- `jest` — **4087 passed, 229 suites, 0 failed**

### Not addressed from the review

Medium 1 (`sendApiMessage`'s hard cap has no test of its own) and Minors 1–4 are still open — say the word and they go in the same branch.

### DIFF
--- a/API.md
+++ b/API.md
@@ -1540,7 +1540,7 @@ Every event carries a `seq` (the turn-scoped sequence number) in addition to the
 | `thinking` | `text` | no | Agent reasoning (if available) |
 | `timeout` | `message`, `resumable: true` | **no** | The soft response budget elapsed, but the turn is still running — see below |
 | `result` | `text`, `request_id`, `session_id`, `duration_ms`, `attachments?` | yes | Final aggregated result; `attachments` present only when images were captured |
-| `error` | `message` | yes | The turn failed |
+| `error` | `message`, `code?` | yes | The turn failed |
 
 The stream ends with `data: [DONE]` after `result`.
 
@@ -1550,6 +1550,20 @@ subprocess is still working and its reply will still be persisted to history.
 Render it as "still working", not as an error. Only `error` is a failure — that
 includes the hard cap (a further 10 minutes), at which point the turn really is
 abandoned.
+
+**Branch on `code`, not on `message`.** An `error` event carries the originating
+failure's code when it has one — `TIMEOUT` for the hard cap, `PROCESS_EXITED`
+for a subprocess crash. The field is omitted when there is no code, so treat it
+as optional. It is present on replayed frames too, so a client that resumed
+through [`GET …/stream`](#resuming-an-interrupted-stream) learns exactly what
+the original connection would have.
+
+**What the hard cap does.** At the cap the turn is *interrupted*, not merely
+abandoned: the subprocess is sent `SIGINT` so it stops working and stops
+consuming tokens. History gets exactly one assistant row for that turn — any
+text the turn had already streamed, followed by `⚠️ Agent response timed out.`
+A hard-capped turn never produces a later reply, so it can never leave both a
+failure row and a reply row behind.
 
 ### Resuming an interrupted stream
 
@@ -1728,7 +1742,7 @@ introducing yourself and what you can help with.
 **Notes:**
 - `GREETING.md` is **deleted before streaming begins**. Subsequent calls return 204 immediately, making the endpoint idempotent. Re-provisioning `GREETING.md` enables a new greeting on the next call.
 - The SSE stream format matches `POST /messages` with `stream: true` — use the same client-side handler.
-- If the agent errors mid-stream, an `{"type":"error","message":"..."}` SSE event is sent and the stream closes.
+- If the agent errors mid-stream, an `{"type":"error","message":"...","code":"..."}` SSE event is sent and the stream closes (`code` omitted when the failure carries none).
 
 ---
 
--- a/API.md
+++ b/API.md
@@ -1531,20 +1531,21 @@ Regardless of `allow_tools`, the agent will not create or update workspace ident
 
 **Event types:**
 
-> **⚠️ Breaking in 1.9.0 — the soft timeout is no longer terminal.**
-> Through 1.8.2 a turn that passed `timeout_ms` emitted a terminal
+> **⚠️ Breaking change — the soft timeout is no longer terminal.**
+> Up to and including 1.8.2, a turn that passed `timeout_ms` emitted a terminal
 > `{"type":"error","message":"Agent response timeout"}` and the stream ended
-> there. From 1.9.0 it emits a **non-terminal** `timeout` event
-> and the connection stays open until the turn finishes or the hard cap (a
-> further 10 minutes) fires.
+> there. It now emits a **non-terminal** `timeout` event instead, and the
+> connection stays open until the turn finishes or the hard cap (a further 10
+> minutes) fires.
 >
-> A pre-1.9.0 client that ignores unknown event types will not fail at
-> `timeout_ms` any more — it will sit on an open connection for up to 10 extra
-> minutes waiting for a terminal frame. **Handle `timeout` explicitly**: render
-> it as "still working", and if you need the old cut-off, close the connection
-> yourself when you receive it (the turn keeps running server-side and its reply
-> is still persisted to history; you can read it back via
-> [`GET …/stream`](#resuming-an-interrupted-stream) or from the session history).
+> A client written against 1.8.2 or earlier that ignores unknown event types
+> will not fail at `timeout_ms` any more — it will sit on an open connection for
+> up to 10 extra minutes waiting for a terminal frame. **Handle `timeout`
+> explicitly**: render it as "still working", and if you need the old cut-off,
+> close the connection yourself when you receive it (the turn keeps running
+> server-side and its reply is still persisted to history; you can read it back
+> via [`GET …/stream`](#resuming-an-interrupted-stream) or from the session
+> history).
 
 Every event carries a `seq` (the turn-scoped sequence number) in addition to the fields below.
 
--- a/API.md
+++ b/API.md
@@ -1509,9 +1509,11 @@ data: [DONE]
 > When images are captured during the turn, the `result` event also includes `"attachments": [{"type":"image","url":"..."}]`.
 ```
 
-> `seq` is the event's position in the turn, counting from 1. Remember the last
-> one you processed — it is the cursor for
-> [resuming an interrupted stream](#resuming-an-interrupted-stream).
+> `seq` is the event's position **within this turn**, counting from 1 — it
+> restarts at 1 on the next turn, so it is a cursor only in combination with the
+> turn's `request_id`. Remember both: they are what
+> [resuming an interrupted stream](#resuming-an-interrupted-stream) takes.
+> (`data: [DONE]` is a stream terminator, not an event, and carries neither.)
 
 ### Requests with tool use
 
@@ -1553,7 +1555,9 @@ Regardless of `allow_tools`, the agent will not create or update workspace ident
 > via [`GET …/stream`](#resuming-an-interrupted-stream) or from the session
 > history).
 
-Every event carries a `seq` (the turn-scoped sequence number) in addition to the fields below.
+Every event in the table below carries a `seq` — the turn-scoped sequence number
+described above — in addition to its own fields. The `data: [DONE]` line that
+closes the stream is a terminator, not an event, and has no fields at all.
 
 | Type | Fields | Terminal | Description |
 |------|--------|----------|-------------|
@@ -1565,6 +1569,13 @@ Every event carries a `seq` (the turn-scoped sequence number) in addition to the
 | `error` | `message`, `code?` | yes | The turn failed |
 
 The stream ends with `data: [DONE]` after `result`.
+
+**Idle streams send a keepalive.** While a turn is working without producing
+events — a long tool call, a slow model — the connection emits an SSE comment
+line (`: keepalive`) every 15 seconds so a reverse proxy does not close it as
+idle. Comment lines are discarded by every conforming SSE parser, carry no
+`seq`, and never reach your event handler; if you parse the stream by hand,
+ignore any line that starts with `:`.
 
 **`timeout` is not a failure.** It means the turn passed `timeout_ms` without
 finishing; the connection stays open and the turn keeps streaming, because the
--- a/API.md
+++ b/API.md
@@ -1638,7 +1638,7 @@ with the same `result` + `[DONE]` frames the original connection would have got.
 | Query param | Description |
 |-------------|-------------|
 | `after_seq` | Resume after this sequence number. Omit (or `0`) to replay the turn from its first event. |
-| `request_id` | The turn you mean to resume — the `request_id` its frames carry. **Required whenever `after_seq > 0`**; optional (and unused) when replaying from the start. |
+| `request_id` | The turn you mean to resume — the `request_id` its frames carry. **Required whenever `after_seq > 0`**; optional when replaying from the start, and the frames you get back report the turn's own `request_id` either way. |
 
 ```bash
 # The stream died after seq 12 — pick the same turn back up.
@@ -1656,6 +1656,10 @@ before seq 12 silently missing. The endpoint answers `400` instead: send the
 `request_id` from the frames you received, or drop `after_seq` and replay the
 current turn from its first event.
 
+**A client that reloaded has no `request_id`, and that is fine.** Drop
+`after_seq`, and the frames you get back carry the turn's own `request_id` —
+which is what you feed to the next resume, cursor and all.
+
 **Resuming is never a conflict.** `409` still means "you tried to start a second
 turn on a busy session"; it is never the answer to a resume. When a turn cannot
 be resumed the endpoint answers `410 Gone` with a `code` saying why:
@@ -1664,15 +1668,15 @@ be resumed the endpoint answers `410 Gone` with a `code` saying why:
 |--------|---------|------------|
 | `TURN_GONE` | No turn for that session — it never ran, or it finished more than 2 minutes ago | Read the session's history |
 | `TURN_MISMATCH` | The session has a turn, but not the `request_id` you asked for — yours is over | Read the session's history |
-| `TURN_TRUNCATED` | That cursor's events have been evicted (see the buffer limits below) | Read the session's history |
+| `TURN_TRUNCATED` | That cursor's events have been evicted (see the buffer limits below) — only a non-zero `after_seq` can get this | Retry **without** `after_seq` |
 | `CURSOR_AHEAD` | `after_seq` is past the turn's last event | Retry **without** `after_seq` |
 
 Each response carries a `hint` field with the same advice.
 
-**`CURSOR_AHEAD` is the one you can recover from without history.** It means the
-named turn is live but has not reached your cursor — re-attaching would match no
-event — so the safe answer is to replay from the start: drop `after_seq` and call
-again.
+**`CURSOR_AHEAD` and `TURN_TRUNCATED` are the two you can recover from without
+history.** Both mean the turn is live and only your *cursor* is unusable — past
+the turn's last event, or pointing into events the buffer has already dropped.
+The answer to either is the same: drop `after_seq` and call again.
 
 A completed turn stays replayable for **2 minutes** after its terminal frame,
 which is what makes a reload immediately after the answer arrives still work. The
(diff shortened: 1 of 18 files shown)


0xMaxMa:
## Review passes

Two review passes over the initial commit (`78bba8c`), then a second independent
pass over the result. Four confirmed defects, all fixed in `0075936` — each one a
distinct way the resume mechanism could strand a client rather than a style nit.

| # | Defect | Consequence | Fix |
|---|---|---|---|
| 1 | `TurnStreamRegistry.complete(sessionId, …)` terminated whichever turn held the session id, not the caller's own | A slow API turn finishing after a channel turn started on the same session id delivered its result to the *newer* turn's client, and scheduled that turn's release | `complete(turn, …)` takes the stream; only the installed record gets a grace timer |
| 2 | `attach()` replaced the sink without telling the old one | An SSE response displaced by a resume hung open forever, waiting on a terminal frame it would never be sent | `TurnSink.displaced?()`; the SSE sink calls `res.end()` |
| 3 | `request_id` fell back to the session id when the client named none | Resuming client sees a plausible request id that correlates with nothing | field omitted instead |
| 4 | `callbackSink.finish` handled only `result` / `error` | Any other terminal frame closed nothing — response open forever | unconditional `else`; message derived from the frame |

Defect 1 is the one worth a second look on merge: `turnStreams` is keyed by
session id, and `sendMessageToSession` (the channel path, reachable via
`POST /v1/agents/:agentId/chats/:chatId/sessions/:sessionId/messages`) shares
that keyspace with `sendApiMessageStream`. Verified by test —
*"a superseded turn finishing late terminates ITSELF, never the turn that
replaced it"*.

### Verification

- Regression proof re-run after the review moved the code: reverting the two
  load-bearing lines (`if (!this.sink) return;` in `emit`, and restoring the
  terminal soft timeout) fails **8 tests** for the right behavioural reasons.
  Restored → green.
- `tsc --noEmit` clean.
- Full suite: **4077 passing, 228 suites**.
- 4 new tests for the defects above (32 across the three new files).

### Not changed

`sendMessageToSession`'s timeout is still terminal — the channel paths have no
resume endpoint to reconnect through, so making it non-terminal there would
drop the turn with nothing able to pick it up. Noted in the PR body as scope.

PR #426, opened 2026-09-01
fix(apps): stop a cold boot-restore from cancelling its build and latching the app off

### LINKED ISSUE #425: [Bug] Boot-time app restore is sized for a warm reboot: a cold rebuild times out, is cancelled, and then latches off permanently
## Summary

The boot-time app restore assumes the machine it wakes up on still has its Docker image cache. When it doesn't — the data directory restored onto a fresh host, a migrated machine, a pruned Docker state — every app has to be rebuilt from source, and the restore's 180 s per-app budget cannot cover that. The timeout then **cancels the build outright** rather than just abandoning the wait, and the resulting "no containers" state is persisted back as `stopped`, so the app is skipped on every future boot as well. One slow first build turns into an app that never starts itself again.

## Investigation

`restoreRunningApps()` (`src/apps/installer.ts:1390`) is fired from boot at `src/index.ts:871`, fire-and-forget. For each app recorded `running` it runs, via `composeUpAsync()` (`src/apps/installer.ts:2918`):

```
docker compose -p <app> up -d --wait      # timeout: RESTORE_COMPOSE_TIMEOUT_MS
```

with `RESTORE_COMPOSE_TIMEOUT_MS = 180_000` (`src/apps/installer.ts:230`) and up to `RESTORE_MAX_CONCURRENCY = 4` (`:234`) apps in flight.

That command is the right one — `up` builds any image that is missing, so a restored app folder is sufficient to bring the app back with no extra machinery. Verified directly: a throwaway compose project with a `build:` service and its image deleted went `Image Built → Container Started → Healthy` from that command alone. The design is sound; the budget and the failure handling are not.

### 1. The 180 s budget is a warm-reboot budget

The comment on `restoreRunningApps()` states the motivation plainly: after a restart the containers are down and the routes point at dead ports. In that scenario 180 s is generous — images exist, `up` just recreates and starts.

On a cold machine the same 180 s has to cover base-image pulls, a full dependency install, an application build, container creation, **and** the `--wait` healthcheck — for up to 4 apps competing for the same CPU. A single app with two `build:` services and a framework build from source will not finish in that window on a modest host.

### 2. A timeout during the build phase cancels the build

`defaultAsyncSpawn()` (`src/apps/installer.ts:3498`) SIGKILLs the CLI on timeout, and its doc comment (`:3486-3497`) argues this is harmless:

> SIGKILL kills the `docker compose` CLI process we spawned, NOT the containers. dockerd keeps bringing them up in the background, so a timeout means "we stopped waiting on the healthcheck", not "the start was cancelled" — the app may well come up healthy moments later.

That holds **after** the images exist. It does not hold during a build: the build session belongs to the client, so killing the CLI cancels it. Verified directly — a compose project whose build takes 30 s, CLI SIGKILLed at 6 s, inspected 50 s later (well past when the build would have finished):

```
image:     <none>
container: <none>
```

No image, no container, nothing coming up in the background. On a cold machine the timeout is not "we stopped waiting", it is "the app was never built".

### 3. The failure latches, so later boots skip the app too

`restoreRunningApps()` only considers apps whose stored status is `running` (`src/apps/installer.ts:1392`). After a cancelled build there are no containers, and:

- `GET /api/v1/apps` (`src/api/apps-router.ts:96-102`) calls `reconcileStatuses()` on every read;
- `queryRuntimeStatus()` (`src/apps/installer.ts:1352`) runs `compose ps` and maps the result through `mapContainerStatesToAppStatus()`, which returns `stopped` for an empty container list (`:3445`);
- `reconcileStatus()` (`:1324`) **persists** that back to the registry (`:1328`).

So the first read of the Apps page after the failed restore rewrites the entry to `stopped`, and from then on `restoreRunningApps()` skips it on every boot. Recovery requires the operator to notice and start the app by hand.

The same mechanism also fires *without* any timeout. `queryRuntimeStatus()` only guards the `building` status (`:1354`); nothing marks an app as "restore in flight". A read that lands while an app's own `compose up` is still working sees zero containers and persists `stopped` underneath it. On a warm reboot that window is seconds; on a cold one it is minutes, so the race is close to guaranteed there.

### 4. The failure is log-only

Failures are collected and emitted as `globalLogger.warn` at `src/index.ts:875-877`. Nothing reaches the registry entry or the API, so an operator looking at the Apps page sees an app that is simply `stopped`, with no indication that the gateway tried to start it and was cut off.

## Proposed Approach

1. **Do not bound the build with the wait budget.** In `composeUpAsync()` (`src/apps/installer.ts:2918`), split the work: run `docker compose -p <app> build` under a separate, much larger budget, then `up -d --wait` under the existing one. This keeps the healthcheck wait short (its original purpose) while letting a cold build take the time it needs. Make both budgets configurable rather than hard-coding a bigger constant at `:230`.

2. **Mark apps as restore-in-flight.** Add an in-memory set on the installer, mirroring the existing `installingNames` guard, populated for the duration of each app's restore in `restoreRunningApps()` (`:1399-1408`) and consulted by `queryRuntimeStatus()` alongside the existing `building` check (`:1354`). While an app is being restored, return the stored status instead of reconciling — so a concurrent read cannot latch it to `stopped`.

3. **Surface the failure.** Record restore failures where `GET /api/v1/apps` can report them (registry entry field, or an installer-held map keyed by app name exposed through `src/api/apps-router.ts:96`), so a failed restore is distinguishable from a deliberately stopped app.

4. **Correct the doc comment** at `src/apps/installer.ts:3486-3497` to state that the "dockerd keeps going" reasoning applies only once the images exist, and that a timeout during a build cancels it.

Optionally: retry a failed restore once, after the first pass drains, so a transient resource crunch during the 4-way parallel start is not terminal.

## Acceptance Criteria

- [ ] A cold restore whose image build exceeds the healthcheck-wait budget still completes: the build is issued under its own, larger budget rather than being killed by the wait timeout.
- [ ] Both budgets are configurable, with the current 180 s preserved as the default for the wait.
- [ ] `queryRuntimeStatus()` returns the stored status, and does not persist `stopped`, for an app whose boot restore is in flight — asserted through the existing spawn seam, with no real Docker.
- [ ] An app that failed its boot restore is not silently indistinguishable from a stopped one: the failure is observable through the apps API, not only the log.
- [ ] An app whose restore failed is not permanently excluded from future boot restores by that failure alone.
- [ ] The `defaultAsyncSpawn()` doc comment no longer claims unqualified that a killed CLI leaves containers coming up.
- [ ] Existing `restoreRunningApps()` tests (`tests/unit/apps/installer.test.ts:822-910`) still pass, extended to cover the build/wait split and the in-flight guard.

## Out of Scope

- Any change to the install/update path's own compose timeouts — this is about the boot-time restore only.
- Changing app container restart policies, or pre-baking/shipping images so no rebuild is needed.
- Reworking how `mapContainerStatesToAppStatus()` classifies container states; the empty-list → `stopped` mapping is correct in isolation and only becomes a problem under the missing in-flight guard.

0xMaxMa:
Closes #425

## Summary

`restoreRunningApps()` brought every app stored as `running` back up with a single `docker compose up -d --wait` under a 180 s budget. That budget assumes the machine still has the app's image cache. When it doesn't — data dir restored onto a fresh host, migrated machine, pruned Docker state — the same 180 s also has to cover base-image pulls, a dependency install and an application build, for up to 4 apps competing for the same CPU.

And the timeout there is not the harmless "we stopped waiting" the code claimed. `defaultAsyncSpawn()` SIGKILLs the compose CLI; once images exist dockerd keeps going, but a **build session is driven by the client**, so killing the CLI cancels it — no image, no container, nothing arriving later. The app then has zero containers, the next status read maps that to `stopped` and **persists** it, and since `restoreRunningApps()` only considers entries stored as `running`, the app is skipped on every future boot as well. One slow first build turned into an app that never started itself again.

## Changes

`src/apps/installer.ts` carries the fix; the rest is wiring and docs.

| Area | Change |
|---|---|
| `composeUpAsync()` | Runs `compose build` under its own budget (default **30 min**) before `up -d --wait` under the existing one (default **180 s**). The healthcheck wait keeps its original short shape; a cold build is no longer bounded by it. |
| Build failure handling | The build is best-effort — its error is swallowed so `up` (which rebuilds what is missing and reports the real error with logs) stays the single authority on success. A project with only `image:` services has nothing to build and exits 0 immediately, so the warm-reboot path pays one extra fast CLI call. |
| Configurability | New `gateway.appRestore.buildTimeoutMs` / `.waitTimeoutMs`. Config is untrusted input and the value reaches `setTimeout`, where `NaN`/`0`/negative all mean *no ceiling* — so anything non-numeric, non-finite or non-positive falls back to the default instead of unbounding the command. |
| In-flight guard | `restoringNames` mirrors the existing `installingNames`; `queryRuntimeStatus()` returns the stored status for an app whose restore is in flight, so a concurrent `GET /api/v1/apps` cannot write `stopped` underneath it. On a cold host that race window is minutes, not seconds. |
| Failure visibility | A failed restore is recorded and surfaced as `restoreError` / `restoreFailedAt` on `GET /api/v1/apps` and `GET /api/v1/apps/:name` — previously the failure existed only as a `globalLogger.warn` at boot. Both fields are absent unless a restore failed, so the response shape is unchanged for existing consumers. |
| No permanent latch | `reconcileStatus()` reports a restore-failed app as `error` (honest, and distinguishable from a deliberate stop) but does **not** persist `running` → `stopped`, so the next boot retries it. The marker clears when the app is next observed `running`, on an explicit start/stop, and on uninstall. |
| Doc comment | `defaultAsyncSpawn()` no longer claims unqualified that a killed CLI leaves containers coming up — it now states that this holds only once the images exist, and that a timeout during a build cancels it. |

## Acceptance criteria

All seven from #425 are covered:

- [x] A cold build that exceeds the wait budget still completes — it runs under its own, larger budget
- [x] Both budgets configurable, 180 s preserved as the wait default
- [x] `queryRuntimeStatus()` returns the stored status and does not persist `stopped` for an in-flight restore — asserted through the existing spawn seam, no real Docker
- [x] A failed restore is observable through the apps API, not only the log
- [x] A failed restore does not permanently exclude the app from future boot restores
- [x] `defaultAsyncSpawn()` doc comment corrected
- [x] Existing `restoreRunningApps()` tests still pass, extended for the build/wait split and the in-flight guard

## Tests

Six new cases in `tests/unit/apps/installer.test.ts`, all through the existing async-spawn seam — no Docker required. The key double, `coldHostAsyncSpawn`, models the real semantics: a command whose work exceeds *its own* budget is killed, and a killed build leaves nothing behind.

**Five of the six fail on the pre-fix code.** Verified by reverting the four behavioural changes in `installer.ts` (build/wait split, in-flight guard, reconcile branch, marker population) and re-running:

```
✕ builds under its own budget before waiting, so a cold rebuild is not killed by the wait timeout
✓ still bounds the build: one that outlasts its own budget fails the restore
✕ takes both budgets from config, ignoring values that would remove the ceiling
✕ does not let a status read latch the app to stopped while its restore is in flight
✕ records a failed restore and keeps the app eligible for the next boot
✕ clears the restore failure once the app is observed running
Tests: 5 failed, 5 passed
```

The one that passes both ways is deliberate: it is the anti-over-correction guard asserting the fix *resized* the build ceiling rather than removing it. The last one is the second guard — a restore marker must not pin an app to `error` forever.

## How to test

```bash
npm run build
npx jest tests/unit/apps/installer.test.ts -t "restoreRunningApps"
```

## Gates

| Gate | Result |
|---|---|
| `tsc --noEmit` | exit 0 |
| `npm test -- --ci` | **225 suites / 4056 tests passing** (4050 before + 6 new) |

## Docs

`API.md` gains the `restoreError` / `restoreFailedAt` fields under `GET /api/v1/apps` and a `gateway.appRestore` paragraph; `config.template.json` gains the block with its defaults.

## Out of scope

Per the issue: install/update compose timeouts, container restart policies, and the `mapContainerStatesToAppStatus()` empty-list → `stopped` mapping (correct in isolation) are untouched.


0xMaxMa:
## 🔍 Code Review — #426 `fix(apps): stop a cold boot-restore from cancelling its build and latching the app off` (self-review)

### 📊 Score: 8.7 / 10.0

| Component | Score |
|---|---|
| Code Quality | 8.5 |
| Correctness | 9.0 |
| Production Readiness | 8.5 |

### 💪 Strengths

- The two-budget split is grounded in a real mechanism, not a guess: `defaultAsyncSpawn` owns the compose build session, so a SIGKILL on an expired wait budget **cancels** the build rather than merely abandoning the wait. The doc comments on `RESTORE_COMPOSE_TIMEOUT_MS` / `RESTORE_BUILD_TIMEOUT_MS` (`src/apps/installer.ts:248-262`) state that distinction instead of restating the code.
- `resolveTimeoutMs` (`src/apps/installer.ts:293`) treats config as untrusted at the boundary that matters — `0`, `-1`, `NaN` and non-numbers all mean "no ceiling at all" to `setTimeout`, so all of them degrade to the default. Covered by an explicit test rather than assumed.
- The in-flight marker fixes the *latching*, not just the symptom: `queryRuntimeStatus` refuses to interpret "no containers yet" as `stopped` while a restore is running, so a `GET /api/v1/apps` landing mid-boot can no longer permanently exclude an app from every future restore.
- Two anti-over-correction tests, not just happy-path ones — the build ceiling still bounds a runaway build, and the in-flight marker is proven to clear once the app is observed running.

1. 🟢 `src/apps/installer.ts:293` — `resolveTimeoutMs` is a third near-copy of the untrusted-numeric-config clamp already duplicated verbatim as `numOr` in `src/agent/dreaming/config.ts:55` and `src/agent/knowledge/config.ts:40`.
   → The correct fix is one shared helper (e.g. `src/utils/config-num.ts`) consumed by all three. **Intentionally left**: `numOr` is module-private in two unrelated subsystems and has different semantics (range-bounded, no positivity requirement, no flooring), so reusing it would mean `src/apps` importing from `src/agent/dreaming` — and de-duplicating the two pre-existing copies is a refactor of files this PR does not touch. Noted for a follow-up rather than expanded into scope here.

### 🔴 Critical (0)

(none)

### 🟡 Medium (1 — found and fixed in-review, `perf(apps): build during boot restore only when an image is missing`)

1. `src/apps/installer.ts:3038` `composeUpAsync()` — **the pre-build ran unconditionally, which is a boot-time regression on the warm path the fix was not meant to touch.** `docker compose up` builds only what is *missing*; `docker compose build` re-runs the build **every time it is called**. So every boot would re-execute every Dockerfile of every app with a `build:` service, `RESTORE_MAX_CONCURRENCY`-at-a-time, before anything started. With `appHousekeeping.buildCachePrune` on by default and `buildCacheMaxAgeHours: 168` (`config.template.json:66-70`), a host that has been up a week still has the *image* but no *build cache* — so that re-run is a from-scratch rebuild in front of a start that needed none.
   → **Fixed**: probe the project's images first and build only when one is genuinely absent.
   ```
   docker compose -p <app> config --images   → the images the project needs
   docker image inspect --format '{{.Id}}' … → one call; non-zero if ANY is missing
   ```
   The probe **fails closed** (`needsImageBuild` returns `false` on every error path): "no signal" means skip the pre-build, which leaves `up` behaving exactly as it did before the build step existed. The opposite default would rebuild apps that only needed starting — the very cost the probe exists to avoid. Verified live against the Docker on this host: `config --images` emits the generated `<project>-<service>` name for a `build:`-only service and exits 0; `image inspect` exits 0 when all arguments are present and 1 when any is missing.

### 🟢 Minor (1)

1. `src/api/apps-router.ts:17-24` — `restoreError` surfaces the last 2000 chars of compose stderr to any authenticated key, not only admins.
   → **Verified acceptable, no change**: `GET /api/v1/apps/jobs/:jobId` (`src/api/apps-router.ts:205`) already returns full compose output at the same auth floor, so this introduces no new exposure ceiling. Both remain behind `createApiAuthMiddleware`.

### ✅ AC Coverage (#425)

- [x] A boot restore on a host with no image cache completes instead of being killed mid-build
- [x] The build gets its own budget, separate from the healthcheck wait
- [x] The wait budget stays short — a hung container is not held for the build budget
- [x] Both budgets are configurable and validated (`gateway.appRestore`)
- [x] A status read landing mid-restore does not persist `stopped`
- [x] A failed restore is visible to the operator (`restoreError` / `restoreFailedAt`)
- [x] A failed app is **not** permanently excluded from future boot restores

### 🧪 Gates

| Gate | Result |
|---|---|
| `tsc --noEmit` | ✅ exit 0 |
| `npm run build` | ✅ exit 0 |
| `jest --ci` | ✅ 225 suites / 4057 tests passing |
| Regression tests fail on pre-fix code | ✅ verified — 5 failed on the reverted fix; the new warm-path guard fails on the unconditional-build code |

The one Medium was a self-inflicted performance regression in the first commit and is fixed in the second, with a regression test proven to fail against the unconditional-build version. The remaining Minor items are deliberate, with the reasoning recorded above.

PR #437, opened 2026-09-02
feat(logs): add `gateway logs` and bound the log directory (#435)

### LINKED ISSUE #435: [Enhancement] Make gateway logs readable from the CLI and bounded on disk
## Summary

The gateway's logs are neither readable nor bounded: there is no CLI command to
view them, and nothing ever rotates or deletes them. Operators must know the
on-disk layout of the log directory and reach for `tail`/`journalctl` by hand,
while the directory grows without limit — measured at **553 MB across 59 files**
on a live host after hours (not months) of uptime, the largest single session log
being **205 MB**.

This issue covers both halves, because they meet at one point: a `--follow`
reader has to survive the file being rotated underneath it.

### Read side — no `logs` command

- The dispatcher (`src/cli/index.ts:66-84`) routes `api`, `gateway`, `service`,
  `update`, `claude`, `doctor`, `debug-bundle`, `agents`, `channels`. There is no
  `logs` command, and `src/cli/commands/` contains no `logs.ts`.
- `gateway` accepts only `start|status|restart|stop`
  (`src/cli/commands/gateway.ts:34-40`), so `claude-gateway gateway logs` exits
  with `Unknown: gateway logs (expected start|status|restart|stop)`.
- A `logs` command was in scope for #374 but was deliberately dropped in #375:
  "accepted but always failing — was removed rather than left as a stub". That
  was the right call at the time, but nothing has tracked re-adding it since.
- The data itself is well structured. `src/logger.ts:19-46` appends one JSON
  object per line to `<logDir>/<agentId>.log`, and `src/index.ts:597` creates the
  `gateway` logger — so `logs/gateway.log` is the process-wide runtime log, with
  per-agent, receiver, and session streams alongside it.
- The path-resolution logic a `logs` command needs already exists, privately, in
  `src/cli/commands/debug-bundle.ts`. `resolveLogDir()` (:48-53) applies the
  `--logDir` -> `config.gateway.logDir` -> default precedence and expands the
  literal `~` that the shipped config template writes; `listSessionLogs()`
  (:64-86) returns the errno instead of reporting every failure as "no logs
  found". Both should be shared, not copied.
- `--follow` has nothing to reuse: `src/shell/tailer.ts` parses Claude Code
  transcript records, not arbitrary files.

### Write side — no rotation or retention

- `src/logger.ts:41-45` is the only write path:
  `fs.appendFileSync(this.logFilePath, line + '\n', 'utf-8')`. There is no size
  check and no cleanup anywhere in `src/` — grepping `rotate|logrotate|maxSize`
  returns only unrelated hits (a CSS keyframe in `src/ui/web-ui.ts`, and
  `src/shell/tailer.ts:172` coping with a file rotated by something else).
- Every level is written unconditionally — `log()` (:25-46) has no level gate —
  and session processes log each stream event at `debug`
  (`src/session/process.ts:287`). A single busy session therefore dominates the
  whole directory.
- The same call also writes a pretty-printed, multi-line copy to stdout (:38),
  which under systemd is duplicated into the journal, so the on-disk figure
  understates the true cost.
- `gateway.logDir` (`src/types.ts:193`) is the only log-related setting. There is
  no retention knob, unlike conversation history, which already has
  `retentionDays` (`src/types.ts:7`, default 60).
- A precedent for the cleanup shape exists: `appInstaller.startBackupCleanup()`
  (`src/index.ts:796-799`, from #310) prunes app backups on a daily unref'd timer
  under a retention-count + max-age union policy.
- Secondary defect in the same function: the append is wrapped in `catch {}` with
  "If we can't write to the log file, just continue" (:43-45). A full disk or a
  permission change silently stops all file logging with no signal anywhere —
  precisely the failure that would make the next incident undiagnosable.

## Proposed Approach

**Read side**

1. Extract `resolveLogDir()` and the directory read out of `debug-bundle.ts` into
   a shared module (e.g. `src/cli/logs-dir.ts`) and have `debug-bundle` import it,
   so tilde expansion and errno reporting cannot drift apart.
2. Add `logs` to the verb switch in `src/cli/commands/gateway.ts`, including its
   usage line in the `writeCommandHelp()` call.
3. Implement:
   `claude-gateway gateway logs [--follow] [--lines <n>] [--agent <id>] [--json]
    [--logDir <path>] [--config <path>]`
   - default: the last 50 lines of `gateway.log`, rendered as
     `<ts> <level> <message>` with `data` appended compactly
   - `--agent <id>`: read `<id>.log` instead
   - `--json`: emit the stored lines verbatim
   - `--follow`: stream appended lines until interrupted
4. Update the `CORE_HELP` row in `src/cli/index.ts:283` to
   `gateway status|restart|stop|logs`.

Every failure must be explicit — missing directory, unknown agent, bad `--lines`
— never an empty result that reads like "no logs".

**Write side**

5. Add a `gateway.logs` config block in `src/types.ts` — `maxFileBytes`,
   `maxFiles` (generations kept), `retentionDays` — named to match the existing
   history-retention settings, with defaults applied when the block is absent.
6. In `src/logger.ts`, check size on append and roll `<name>.log` ->
   `<name>.log.1` ..., dropping generations past `maxFiles`. Best-effort and
   synchronous, so logging still cannot throw into a caller.
7. Add a boot-time and daily sweep for age-based deletion, modelled on
   `startBackupCleanup()` in `src/index.ts` (unref'd timer).
8. Gate `debug` writes behind a configurable level so one session cannot dominate
   the directory.
9. Surface a write failure at least once (latched per process) instead of the
   bare `catch {}`.

**Where the two halves meet**

10. `--follow` must detect that the file it holds has been rotated away and
    reopen the new one, rather than silently following a deleted inode.

## Acceptance Criteria

- [ ] `claude-gateway gateway logs` prints the tail of `gateway.log` and exits 0
- [ ] `--agent <id>` reads `<id>.log`; an unknown id exits non-zero, naming the
      path tried and the ids available
- [ ] `--lines <n>` is validated as a positive integer; bad values exit non-zero
- [ ] `--follow` streams appended lines and exits 0 on SIGINT
- [ ] `--json` output is byte-identical to the stored lines
- [ ] A `~` in `gateway.logDir` is expanded; an unreadable directory reports the
      errno instead of an empty listing
- [ ] `gateway --help` and the top-level help both list the verb
- [ ] `debug-bundle` still works, off the shared helper
- [ ] A file exceeding `maxFileBytes` is rotated and a new file started
- [ ] At most `maxFiles` generations are kept per stream; the oldest is deleted
- [ ] Files older than `retentionDays` are removed at boot and on the daily sweep
- [ ] A config with no `gateway.logs` block keeps working on defaults
- [ ] Rotation never throws into a logging caller, and a write failure is
      surfaced at least once rather than silently swallowed
- [ ] `--follow` reopens the file across a rotation instead of following the old
      inode
- [ ] Unit tests cover the size threshold, generation pruning, age pruning, and
      the CLI flag validation

## Out of Scope

- Redaction of `logs` output. `debug-bundle` redacts because its output is meant
  to be shared; `logs` prints local files the operator can already `cat`. A
  `--redact` flag can be a follow-up.
- Selecting a specific session log (`--session`), and reading logs from a remote
  host.
- Journal/stdout volume, which the operator controls via `SystemMaxUse`.
- Compressing rotated files.

0xMaxMa:
Closes #435

## What the measurement changed

The issue asks for two halves — a `gateway logs` reader and a bounded writer. Measuring the live log directory first reordered them.

| | |
|---|---|
| Log directory | **557 MB across 59 files** (the issue said 553 MB; it grew during the check) |
| Largest single file | **217 MB** |
| Session logs | **99% of the total** |
| Everything else | **6.1 MB across ~40 files**; `gateway.log` is 31 KB |
| Level mix, first 20k lines of the largest file | **19,995 `debug` : 5 `info`** |

Every session stream event is logged at `debug` (`src/session/process.ts:900`). So the level gate, not rotation, is the fix that matters here: rotation alone would have gone on writing all 551 MB and merely deleted it sooner. **Rotation and retention bound what is *kept*; only the level bounds what is *written*.**

`level` defaults to `info`. This does not affect `debug-bundle`, which selects on warn/error and `submit-diag` — and `submit-diag` is emitted via `logWarn` (`src/shell/claude-pty-shell.ts:604`), so it still passes the gate.

## Write side — `src/logger.ts`

New optional `gateway.logs` config block; omit it entirely and the defaults apply.

| Field | Default |
|---|---|
| `level` | `info` |
| `maxFileBytes` | 16 MiB |
| `maxFiles` | `3` |
| `retentionDays` | `14` |

- **Level gate** on the file *and* stdout. stdout is duplicated into the systemd journal, so gating only the file would pay for the volume twice.
- **Size rotation**, `<name>.log` → `.log.1` → `.log.N`. Lowering `maxFiles` collects the generations it orphans — `renameSync` clobbers its destination, so a fixed cap looks self-enforcing right up until the config changes underneath it.
- **Age sweep** at boot and once a day, modelled on `startBackupCleanup()`. Age rather than count, because each session writes its own `<agent>:session:<uuid>.log` and never returns to it, so a per-stream generation cap can never reach them. The timer is unref'd.
- **Write failures surface.** The append was wrapped in a bare `catch {}`: a full disk or a permission change silently stopped all file logging, which is precisely the failure that makes the next incident undiagnosable. Now latched and reported once to stderr — reporting every occurrence would be its own denial of service, since the failing call *is* the logger.

Two decisions worth flagging for review:

- **Policy is module state, not a per-logger argument.** There is one policy per process and `createLogger()` has 13 call sites, several of which (routers, the session process) have no access to the gateway config. Threading an options bag through all 13 would put the same value in 13 places and let them drift. `configureLogging()` is called in `src/index.ts` before the first logger exists — verified to be the only process entry point (`dist/entry.js`), so it covers the session loggers too.
- **The size counter is shared per absolute path.** One file legitimately has several logger instances — `createLogger(agentConfig.id, …)` runs from both the boot path and the runner constructor — and per-instance counters each see only their own share and rotate late.

## Read side

```
claude-gateway gateway logs [--follow] [--lines <n>] [--agent <id>] [--json] [--logDir <path>] [--config <path>]
```

Reads the files directly, so it answers whether or not the gateway is running — which is usually exactly when you need it.

- `resolveLogDir()` and the directory listing move out of `debug-bundle` into `src/cli/logs-dir.ts`. Two copies would drift, and the tilde bug is what a drifted copy looks like: the shipped config template writes a literal `~`, and `readdirSync('~/…')` throws ENOENT — which reads exactly like "the directory is empty".
- **Every failure is explicit.** An unknown `--agent` names the path it tried and lists the ids that do exist; an unreadable directory reports its errno. A log reader must never answer "no logs" when the real answer is "I looked in the wrong place".
- **`--follow` polls and compares the inode** rather than using `fs.watch`. This is where the two halves of the issue meet: the reader holds a file the writer is free to rename out from under it, and a watcher would hold the dead inode and go quiet forever while the gateway wrote happily into the new file.
- `--follow` is registered as a *global* boolean flag because the verb is a positional — `gateway --follow logs` would otherwise consume `logs` as the flag's value and leave no command at all. No generated resource command declares a `--follow` value flag (verified).

`--json` prints the stored lines verbatim. Note this output is **not** redacted, unlike `debug-bundle` — it is the local file you could already `cat`. The README says so where the command is documented.

## Tests

48 new tests across `tests/unit/logger-rotation.test.ts`, `tests/unit/cli-logs.test.ts`, `tests/unit/cli-logs-dir.test.ts`.

Each of the **12 guards was revert-proven** — patched back out, confirmed its tests fail for the right reason, restored. Two tests did not survive that and were strengthened:

- `maxFiles: 0` passed with rotation entirely removed, because it only checked that the live file contained the newest line. It now asserts the file *restarted* — otherwise "keep no generations" is indistinguishable from "never rotate", which is unbounded growth.
- The shared-counter test sampled the size only at the end. A late-rotating writer overshoots (423 bytes against a 400-byte threshold, measured) and then rotates back under it, so the final size alone cannot tell the two apart. It now samples after every write.

Full suite: **235 suites / 4277 tests passing** (baseline 232 / 4219). `npx tsc --noEmit` clean and `npm run gen-cli:check` reports the generated files up to date. There is no `lint` script in this repo, so no lint gate was run — an earlier version of this line claimed one, which was wrong.

## Note on the issue text

#435 cites `src/session/process.ts:287` as the per-stream-event `debug` site. That line is the `createLogger` call; the actual site is `src/session/process.ts:900`. The claim itself is correct — only the line number is off.

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,6 +305,10 @@ only the level bounds what is *written*.
 Retention is age-based because each session writes its own `<agent>:session:<uuid>.log` and never
 returns to it — `maxFiles` prunes generations of one stream, so it can never reach them.
 
+This block is **hot-reloaded**: edit it in `config.json` and it applies on the next config reload,
+no restart. That matters because turning the level up is something you do while chasing a live
+problem, and a restart would kill the sessions you are trying to observe.
+
 ### `session`
 
 | Field | Default | Description |
--- a/README.md
+++ b/README.md
@@ -135,6 +135,7 @@ Once installed, drive it through the CLI — it detects whichever manager owns t
 claude-gateway gateway status   # manager, URL, health
 claude-gateway gateway restart
 claude-gateway gateway stop
+claude-gateway gateway logs     # tail the gateway's own log (works even when it is dead)
 ```
 
 Managing PM2 directly still works too:
@@ -214,6 +215,12 @@ Config lives at `~/.claude-gateway/config.json` (or set `GATEWAY_CONFIG` env var
   "configVersion": "1.0.0",
   "gateway": {
     "logDir": "~/.claude-gateway/logs",
+    "logs": {
+      "level": "info",
+      "maxFileBytes": 16777216,
+      "maxFiles": 3,
+      "retentionDays": 14
+    },
     "timezone": "Asia/Bangkok",
     "api": {
       "keys": [
@@ -276,6 +283,27 @@ returns the `token` (the share endpoint stays enabled) and callers with their ow
 public base — e.g. LINE, which derives its host from the inbound webhook — build
 `<base>/shared/<token>` themselves. HTTP is accepted only for local development
 hosts such as `http://host.docker.internal:10850/gateway`.
+
+### `gateway.logs` (optional)
+
+Verbosity, rotation and retention for the files in `logDir`. The whole block is optional —
+omit it and the defaults below apply.
+
+| Field | Default | Description |
+|-------|---------|-------------|
+| `level` | `"info"` | Minimum level written, to both the file and stdout. One of `debug`, `info`, `warn`, `error` |
+| `maxFileBytes` | `16777216` (16 MiB) | Rotate `<name>.log` to `<name>.log.1` once an append would carry it past this size |
+| `maxFiles` | `3` | Rotated generations kept per stream; the oldest is deleted. Lowering it collects the generations it orphans at the next rotation. `0` = keep none |
+| `retentionDays` | `14` | Delete logs (live and rotated) older than this, at boot and once a day. `0` = keep forever |
+
+`level` is the one that governs disk usage. Session processes log every stream event at `debug`,
+which on a live host measured 19,995 `debug` lines to 5 `info` lines inside a single 217 MB file —
+so `debug` is off by default. Set `"level": "debug"` when you are actually chasing something, and
+expect the directory to grow quickly while it is on. Rotation and retention bound what is *kept*;
+only the level bounds what is *written*.
+
+Retention is age-based because each session writes its own `<agent>:session:<uuid>.log` and never
+returns to it — `maxFiles` prunes generations of one stream, so it can never reach them.
 
 ### `session`
 
@@ -715,6 +743,7 @@ The `claude-gateway` binary doubles as a command-line client for a running gatew
 claude-gateway                             # help (never starts a server)
 claude-gateway gateway start               # run the gateway in the foreground
 claude-gateway gateway status              # is it running? which manager owns it?
+claude-gateway gateway logs --follow       # stream the gateway log (reads files, needs no server)
 claude-gateway service install             # run it as a systemd-user (or --manager pm2) service
 claude-gateway update check                # newer claude-gateway published?
 claude-gateway claude update               # update Claude Code via its own updater
@@ -726,6 +755,28 @@ claude-gateway crons run <jobId>
 claude-gateway debug-bundle                # small redacted bundle for a stuck session (works even if the server is down)
 claude-gateway api GET /v1/agents          # escape hatch: call any endpoint directly
 ```
+
+### Reading the logs
+
+`gateway logs` reads the log files directly, so it answers whether or not the gateway is
+running — which is usually exactly when you need it.
+
+```bash
+claude-gateway gateway logs                       # last 50 lines of logs/gateway.log
+claude-gateway gateway logs --lines 200 --follow  # more history, then stream
+claude-gateway gateway logs --agent alfred        # that agent's stream instead
+claude-gateway gateway logs --json                # the stored JSON lines, verbatim
+```
+
+Each line is stored as one JSON object and rendered as `<ts> <LEVEL> <message>` with `data`
+appended; `--json` prints the stored line unchanged, for piping into `jq`. `--agent <id>` takes
+any stream id in the log directory — agents (`alfred`), receivers (`alfred:receiver`), and
+sessions (`alfred:session:<uuid>`) each get their own file — and an unknown id lists the ids that
+do exist rather than reporting an empty result. `--follow` survives a rotation: when the file it
+is watching is renamed away it reopens the new one instead of going quiet.
+
+Unlike `debug-bundle`, this output is **not redacted** — it is the local file you could already
+`cat`. Skim before pasting it anywhere.
 
 > **Upgrading from < 1.8:** a service unit that runs the binary with no command still starts the
 > gateway, with a deprecation warning. Point `ExecStart` at `claude-gateway gateway start`, or
@@ -961,6 +1012,7 @@ claude-gateway/
 ├── config.json                         ← gateway config
 ├── logs/
 │   ├── alfred.log
+│   ├── alfred.log.1                    ← rotated generation (see `gateway.logs`)
 │   └── warrior.log
 ├── shared-skills/                      ← shared skills (synced to ~/.claude/skills/ on boot and on change)
 │   └── <skill-name>/
--- a/src/cli/commands/debug-bundle.ts
+++ b/src/cli/commands/debug-bundle.ts
@@ -2,7 +2,7 @@ import { execSync } from 'child_process';
 import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
-import { expandHome, loadCliConfig } from '../http-client';
+import { listSessionLogs, resolveLogDir } from '../logs-dir';
 import { redactLine } from '../redact';
 import { writeCommandHelp } from '../output';
 
@@ -39,51 +39,6 @@ export function selectDiagnosticLines(content: string, maxLines = MAX_DIAG_LINES
     }
   }
   return picked.length > maxLines ? picked.slice(picked.length - maxLines) : picked;
-}
-
-/** The config file stores `logDir` exactly as written, and the shipped template
- *  writes `~/.claude-gateway/logs`. The server expands that itself on boot, so
- *  the literal tilde survives on disk and only bites a reader that forgets to
- *  expand it — `readdirSync('~/...')` throws ENOENT and looks like "no logs". */
-function resolveLogDir(flags: Record<string, string | boolean>): string {
-  if (typeof flags.logDir === 'string') return expandHome(flags.logDir);
-  const cfg = loadCliConfig(typeof flags.config === 'string' ? flags.config : undefined);
-  if (cfg.logDir) return expandHome(cfg.logDir);
-  return path.join(os.homedir(), '.claude-gateway', 'logs');
-}
-
-/**
- * Session logs in `dir`, or the reason the directory could not be read.
- *
- * The read used to be wrapped in a bare `catch { return [] }`, which reported
- * every failure as "no session logs found". That is how the unexpanded `~`
- * stayed invisible: an ENOENT on a path that was never resolved looked exactly
- * like an empty directory, and a permission problem still would. The caller
- * needs the errno to say anything useful, so it is returned rather than eaten.
- */
-function listSessionLogs(dir: string): { logs: Array<{ file: string; mtimeMs: number }>; error?: string } {
-  let names: string[];
-  try {
-    names = fs.readdirSync(dir);
-  } catch (err) {
-    const e = err as NodeJS.ErrnoException;
-    if (e.code === 'ENOENT') return { logs: [], error: `${dir} does not exist` };
-    if (e.code === 'ENOTDIR') return { logs: [], error: `${dir} is not a directory` };
-    return { logs: [], error: `${dir} could not be read (${e.code ?? e.message})` };
-  }
-  const logs = names
-    .filter((n) => n.endsWith('.log') && /session/i.test(n))
-    .map((n) => {
-      const file = path.join(dir, n);
-      let mtimeMs = 0;
-      try {
-        mtimeMs = fs.statSync(file).mtimeMs;
-      } catch {
-        /* ignore */
-      }
-      return { file, mtimeMs };
-    });
-  return { logs };
 }
 
 function claudeCodeVersion(): string {
--- a/src/cli/commands/debug-bundle.ts
+++ b/src/cli/commands/debug-bundle.ts
@@ -2,7 +2,7 @@ import { execSync } from 'child_process';
 import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
-import { listSessionLogs, resolveLogDir } from '../logs-dir';
+import { explicitConfigWarning, listSessionLogs, resolveLogDir } from '../logs-dir';
 import { redactLine } from '../redact';
 import { writeCommandHelp } from '../output';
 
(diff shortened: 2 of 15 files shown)


0xMaxMa:
## Self-review of this branch (`92f5410`)

Reviewed the full diff against `main` for correctness, edge cases, security, necessity, duplication and consistency with the surrounding code. Five things came out of it; three were real, one was a note, and one turned out not to be a bug.

### Fixed

**🟠 `gateway.logs` was a silent no-op on reload.** The watcher only diffs the gateway fields it names explicitly, and this block was in neither `gatewayFieldPairs` nor `HOT_RELOADABLE_GATEWAY_FIELDS`. Editing `level` in `config.json` therefore produced no change event at all — the new policy was not applied, *and* no "requires restart" warning was emitted either, so the operator got no signal in either direction. Both halves were needed: the pair to detect it, the field list to mark it hot.

Hot rather than cold is the right call here and not a shortcut. Policy is process-wide module state, so re-installing it is one call — and turning the level up is something you do while chasing a live problem, which is exactly when a restart is unaffordable, because it kills the sessions being investigated.

The reload confirmation logs at `warn`, not `info`, deliberately: raising the level to `warn` would otherwise swallow the confirmation under the very policy just installed, leaving no evidence the edit took.

**🟠 `--follow` skipped bytes written between the tail and the follow start.** `runGatewayLogs` printed the tail, then took a fresh `statSync` to decide where to follow from. Anything the gateway appended between those two calls sat past the tail's read and before the follower's start, so nothing ever printed it. Nothing can interleave *within this process* — both calls are synchronous — but the writer is a different process, which is the entire premise of the command. `tailFrom()` now returns the offset it read to and `--follow` resumes from that; `tailLines` is a thin wrapper over it.

**🟡 `retentionDays: 0` at boot meant the sweep timer never existed.** `startLogRetentionSweep` returned early on `retentionDays <= 0`. Since `retentionDays` is read on each run rather than captured, a reloaded policy is supposed to take effect at the next sweep — but a timer that was never created cannot do that, so retention stayed off until a restart no matter what the config said. The timer is now created unconditionally; a disabled sweep costs one no-op call a day.

**🟡 `pruneGenerations` replaces what was dead code.** The old `fs.rmSync(\`${file}.${maxFiles}\`)` never removed anything: `renameSync` clobbers its destination, so the top generation is overwritten by the shift below it anyway. That holds only while `maxFiles` stays put — lower it (5 → 2) and generations 2..5 are orphaned, with nothing renaming onto them and only the age sweep to collect them. Now a readdir-based prune of every generation ≥ `maxFiles`, run once per rotation.

### Noted, not changed

**🔵 `fileSizes` growth.** One entry per log file, and session logs create a new file each. The retention sweep deletes the entry alongside the file, which bounds the map for any config that keeps retention on; with `retentionDays: 0` it grows at roughly a path string per session. Documented at the declaration rather than given a second eviction mechanism whose only job would be to duplicate the sweep.

### Investigated and dismissed

**`tailLines` chunk-boundary off-by-one.** I suspected the reverse reader dropped a real line when a 64 KB chunk boundary landed exactly on a newline, since it unconditionally shifts the leading fragment. It does not: the loop only stops with *more* than `count` newlines in hand, so the fragment dropped is always older than the newest `count`. Verified across alignments either side of 1024 and 2048 lines before concluding — I was one step from "fixing" correct code. The probe is kept as a property test (U-LOGS-20) rather than discarded.

### Tests

- `U-CW-12` / `U-CW-13` — `gateway.logs` is diffed, emitted with `agentId: ''` and `hotReloadable: true`; an identical block produces no change.
- `U-LOG-20` — lowering `maxFiles` collects the orphaned generations.
- `U-LOG-21` — the sweep timer is scheduled even when retention starts at 0, and a later `configureLogging` raises it into effect.
- `U-LOGS-20` — tail is exact at every chunk alignment (4096 × 64-byte lines, `n` swept across the 1024/2048 boundaries).
- `U-LOGS-21` — `endPos` is the offset read to, and does not move when the file later grows.

All 15 guards on this branch are revert-proven: each was re-run against a faithful revert of the code it covers and confirmed to fail. That harness caught two of my own tests that passed with the fix removed (`U-LOG-08` asserted growth rather than truncation; `U-LOG-09` asserted a continuous invariant at a single point in time, which a late-rotating writer satisfies while overshooting) and both were rewritten.

**One limitation, stated plainly:** the `--follow` resume offset cannot be proven by a single-process test — only a second process can open the window it closes, and Node's synchronous `statSync`/`readSync` cannot be made to interleave from inside one. `U-LOGS-21` pins the helper's contract so the offset cannot drift silently, but the wiring itself is correct by construction rather than by a timing test.

### Gates

`npx tsc --noEmit` exit 0. `npm run gen-cli:check` reports generated files up to date. Full suite **235 suites / 4272 tests, 0 failures**.

### Docs

No HTTP API surface is touched, so `API.md` is unaffected. `README.md` is updated in this branch: `gateway logs` in both CLI blocks, a "Reading the logs" subsection (including the warning that this output is **not** redacted, unlike `debug-bundle`), the `gateway.logs` config table, the rotated-file line in the runtime-data tree, and a note that the block is hot-reloaded.

0xMaxMa:
## Second review pass — independent of the first (`0a6caf0`)

The first pass looked at the logger internals. This one deliberately looked elsewhere: the CLI wiring, the boot path, and the seams between the new module and the code that already existed. Three findings, all real, all fixed.

**🟠 `--follow` stopped being a typo everywhere else.** The flag has to be a *global* boolean, because `gateway logs` puts its verb in a positional — `gateway --follow logs` would otherwise consume `logs` as the flag's value and leave no command at all. But `GLOBAL_FLAG_NAMES` was written as a spread of `GLOBAL_BOOLEAN_FLAGS`, so making `--follow` parse everywhere also made it *accepted* everywhere. `claude-gateway crons list --follow` went from "Unknown flag(s): --follow" (exit 1) to silently ignored (exit 0) — which is precisely the failure the unknown-flag check exists to close, and it was reintroduced by a branch that never meant to touch that command. The two sets answer different questions ("how does this parse?" vs "is this accepted here?") and are now written out separately.

**🟠 An unusable `--config` resolved to the default log directory in silence.** `loadCliConfig` answers `{}` for a file that is missing, unreadable or malformed — it cannot distinguish them, and for most commands that shape is right. For a log reader it is not: `resolveLogDir` then quietly returns the *default* directory and the command reports "No log file at `<default>/gateway.log`", naming a directory the operator never asked about while the typo in `--config` goes unmentioned. That is the same class of failure as the unexpanded `~` this PR already fixes, arriving by a different route — and #435 is explicit that a failure must never read as "no logs". `explicitConfigWarning()` now says so on stderr, in both `gateway logs` and `debug-bundle`.

Scoped deliberately: only an *explicit* `--config` is worth complaining about, since a fresh install has no config file and the default directory is the correct answer for it. `--logDir` suppresses the warning as well, because it wins outright and the config was never consulted.

**🟡 The retention sweep misreported its own threshold after a reload.** The `Swept expired log files` line took `retentionDays` from the policy captured at boot. Since `gateway.logs` is now hot-reloadable, a reload left the sweep using the new threshold while logging the old one — a misleading number in the one place added specifically to make this diagnosable. It reads the live policy now.

### Also checked, nothing to change

- **`--agent` traversal.** `/`, `\` and a `..` segment are all rejected; a colon-bearing session id (`<agent>:session:<uuid>`) still works, which is the reason the check rejects traversal rather than restricting the charset. An absolute path and a NUL byte both fail closed.
- **Rotation reopen semantics in `--follow`.** Bytes written to the old inode after the last poll and before the rename are not printed; they are in `<name>.log.1` and the code says so at the point it makes that choice. Reading them would mean following a file by identity across a rename, which is what the poll-and-compare design exists to avoid.
- **`listLogStreamIds`** excludes rotated generations — offering `.log.1` as an `--agent` value would name a target rotation can rename away at any moment.
- **`tailFrom` on an empty file, a non-JSON line, a JSON array line.** All degrade to passing the bytes through rather than dropping them.

### Tests

`U-RC-435a`, `U-LD-10`, `U-LD-11`, `U-LD-12`, `U-LOGS-22` — all three fixes revert-proven (reverting each makes exactly its guards fail, and the revert compiles).

Two limitations stated rather than papered over:

- The stale-`retentionDays` fix has no test of its own. It lives in `main()`, which has no harness in this repo. What it depends on — `loggingConfig()` reflecting a reload — is covered by `U-LOG-21`.
- `U-LOGS-22` needs a hermetic home directory and takes it from a module mock. `jest.spyOn(os, 'homedir')` throws `Cannot redefine property` on this Node, and jest hands each test file a *copy* of `process.env`, so reassigning `HOME` leaves the native lookup reading the real one. Without the mock, the test would read whatever logs happen to exist on the machine running it — passing here and failing on a clean box, or vice versa.

### Gates

`npx tsc --noEmit` exit 0. `npm run gen-cli:check` up to date. Full suite **235 suites / 4277 tests, 0 failures**.

PR #438, opened 2026-09-02
fix(pty): detect the bypass dialog by structure, not screen position (#436)

### LINKED ISSUE #436: [Bug] Bypass Permissions dialog is never detected when it renders at the top of the screen, wedging session startup
## Summary

`ScreenModel.detectDialog()` only scans the bottom 20 rows of the 50-row PTY screen. Claude Code's "Bypass Permissions mode" dialog renders at the **top** of the screen when it appears on a fresh boot — before any conversation output exists to push it down — so both detection markers fall outside the scanned window, `detectDialog()` returns `null`, and `Driver.maybeHandleDialog()` returns without acting or logging.

Nothing is ever pressed. The session sits at the dialog until `STARTUP_TIMEOUT_MS` (120 s) expires, shuts down, and respawns into the same dialog.

This is **not** the bug fixed by #432. #432 corrected *which keystroke* is sent once the dialog is detected, and that fix is correct. This issue is that on a fresh boot the dialog is never detected in the first place, so the #432 code path never runs. #432 is necessary but not sufficient; both are needed.

### Live evidence

Filtering every log file down to genuine `session stderr` entries (an agent transcript that merely quotes the source produces false matches on a naive grep) yields **10 real startup timeouts across 3 different agent sessions**:

```
2026-09-01 11:46 / 11:48                    session A
2026-09-01 11:46 / 11:48 / 11:50            session B
2026-09-01 14:38 / 14:40 / 14:42            session A
2026-09-02 06:30:54                         session A   <- after v1.8.3 (#432 merged)
2026-09-02 06:31:17                         session C   <- after v1.8.3 (#432 merged)
```

The last two are the decisive ones: both are on a build that already contains #432, on two independent sessions.

### The captured screen

`claude-pty-shell.ts:808` dumps `screen.text()` on timeout. The 2026-09-02 06:30:54 dump, row by row (50-row screen):

```
  1  (blank)
  2  ─────────────────────────────────────────  (rule)
  3    WARNING: Claude Code running in Bypass Permissions mode
  4  (blank)
  5    In Bypass Permissions mode, Claude Code will not ask for your approval ...
  6    This mode should only be used in a sandboxed container/VM ...
  7  (blank)
  8    By proceeding, you accept all responsibility for actions taken while running
       in Bypass Permissions mode.
  9  (blank)
 10    https://code.claude.com/docs/en/security
 11  (blank)
 12    ❯ No, exit
 13      Yes, I accept
 14  (blank)
 15    Enter to confirm · Esc to cancel
 16-50  (blank)
```

### The causal chain

| Fact | Source |
| --- | --- |
| PTY screen is 50 rows | `src/shell/screen.ts:381` |
| `detectDialog()` scans `bottomText(DIALOG_REGION_ROWS)` = rows 31-50 | `src/shell/screen.ts:55`, `:487`, `:491-502` |
| Both markers (`Bypass Permissions mode`, `Yes, I accept`) sit at rows 3/5/8 and 13 | dump above |
| Rows 31-50 are entirely blank | dump above |
| `detectDialog()` therefore returns `null` | `screen.ts:498` |
| `maybeHandleDialog()` returns silently on `null` — no log at any level | `src/shell/claude-pty-shell.ts:1040-1048` |
| Startup timeout fires at 120 s, `shutdown(1)`, respawn | `claude-pty-shell.ts:77`, `:807-810` |

### Proof the handler never ran

`logWarn()` (`claude-pty-shell.ts:96-98`) writes to stderr **unconditionally** — there is no `DEBUG` gate on it, unlike `logDebug()`. Every branch of `maybeHandleDialog()` that reaches `decideBypassDialogAction()` emits one of:

- `Bypass Permissions dialog left alone — <reason>`
- `Bypass Permissions dialog: moving caret to the accept row (n/60)`
- `accepting Bypass Permissions dialog ... (n/60)`

The session log for the timed-out session contains **zero** of these lines. The handler was never reached, which can only happen via the `if (!dialog) return` branch.

(Note for anyone reproducing this: grepping a session log for those strings gives false positives, because an agent's own transcript quotes the source code. Filter to entries whose `message` is `session stderr`.)

### Why it is intermittent

`screen.ts:46-53` documents the assumption behind the window:

> A modal dialog is anchored to the bottom; this window comfortably covers a tall dialog box while excluding the upper ~60% of the buffer where conversation/scrollback lives. [...] Bump this if Claude Code's dialog grows.

The comment anticipated the dialog growing *taller*. What actually happens is that it renders at the **top**: this dialog appears at boot, before any conversation exists, so there is no scrollback to push it down. A boot that happens to emit some output first lands the dialog inside rows 31-50 and detection works; a clean boot leaves it at row 3 and detection silently fails. That is the whole source of the intermittency.

### Why the fix is not "make the window bigger"

Widening the window toward the full screen removes the only defence against a false positive: a chat reply, or re-injected conversation history, that quotes the dialog's text. That text is real characters on a real screen, and a false accept would send `Enter` — which, if the caret happens to sit on `No, exit`, terminates Claude Code unrecoverably.

The window is a *proxy* for "this is the live modal, not quoted text". A better discriminator is available in the dump above, and it is position-independent:

1. The two option labels appear as **whole rows** (`No, exit`, `Yes, I accept`), not embedded in prose.
2. Exactly **one** of those two rows carries the `❯` caret.
3. The dialog's own confirm affordance, `Enter to confirm`, renders **below** the option rows.
4. **Nothing renders below that footer** — the modal owns the screen; every remaining row is blank.

Property 4 is the strong one. It holds for the real capture (rows 16-50 blank) and it is false for every quoting scenario, because quoted text always has the input box, its border and the status bar rendered underneath it. It is also what the existing regression fixture for "quoted dialog in scrollback" actually encodes, once you stop reading it as a statement about position.

This is the same shape as the existing `detectRequestTooLarge()` guard, which requires the overlay's own dismiss footer (`TUI_REQUEST_TOO_LARGE_DISMISS`) precisely because the error prefix alone appears in ordinary prose.

### Secondary finding: the failure is silent

`maybeHandleDialog()` logs nothing when detection returns `null`. That is why this took two investigations to find: the symptom (a startup timeout) is three layers away from the cause (a detector that returned `null`), with no breadcrumb in between. When the detection markers are on screen but the structure does not validate, that is worth exactly one warning line.

## Proposed approach

1. Replace the positional window in `detectDialog()` with a structural, whole-screen check implemented in `src/shell/bypass-dialog.ts` (which already owns the option-row patterns and is dependency-free), requiring all of: both markers present, both option rows parseable as whole rows, exactly one caret across them, the `Enter to confirm` footer below them, and only blank rows below the footer.
2. Make `dialogText()` return the same region detection fired on, so `decideBypassDialogAction()` keeps reading exactly what was matched (the existing invariant documented at `screen.ts:481-486`).
3. Apply the same structural precondition inside `decideBypassDialogAction()`, so the action layer independently refuses to act on a screen that is not a live dialog (defence in depth — the two layers must not be able to drift).
4. Emit one warning when the bypass markers are on screen but the dialog does not validate, so a future detection failure is visible instead of silent.

## Acceptance criteria

- [ ] A regression test built from the verbatim 50-row capture above (dialog at rows 2-15, rows 16-50 blank) makes `detectDialog()` return `'bypass-permissions'`, and it fails on the pre-fix code.
- [ ] On that same screen, `decideBypassDialogAction()` returns a `move` toward the accept row (caret starts on `No, exit`).
- [ ] A dialog rendered at the bottom of the screen is still detected (no positional regression in either direction).
- [ ] Both Claude Code renderings still work: numbered rows (`❯ 1. No, exit` / `2. Yes, I accept`) and un-numbered rows.
- [ ] A dialog that has scrolled up into the scrollback, with conversation lines and an idle input prompt rendered below it, is **not** detected and attracts **no** keystroke.
- [ ] Prose that quotes the labels mid-sentence is not detected.
- [ ] A screen with no caret on either option row, or a caret on both, is not detected.
- [ ] A screen whose option rows and footer are present but which has non-blank content below the footer is not detected.
- [ ] `decideBypassDialogAction()` returns `wait` for any screen that would not have been detected, when called directly.
- [ ] The keystroke budget (`BYPASS_MAX_KEYS`) and the miss-tolerance reset (`BYPASS_RESET_AFTER_MISSES`) behave exactly as before.
- [ ] One warning is logged when the markers are present but the dialog does not validate.
- [ ] Comments that describe detection as region-restricted are corrected wherever they appear (`screen.ts`, `bypass-dialog.ts`, `claude-pty-shell.ts`).
- [ ] `npm run typecheck` clean and `npm test` at 100%.

## Out of scope

- Changing `DIALOG_REGION_ROWS` itself — `parseInteractivePrompt()` still uses it for the permission-prompt bottom region, which is a different detector with a different liveness proof (the behavioural probe).
- The `skipDangerousModePermissionPrompt` setting in the user's Claude Code config. It stops the dialog from appearing at all, but it is a machine-global file the gateway does not own, and it would only mask the detector bug rather than fix it.
- Widening `neutralizeTuiTriggers()` to defang the bypass dialog's text in re-injected history. With the "nothing below the footer" rule, re-injected history cannot reproduce a detectable dialog (it always renders inside the input box, with the box border and status bar below it), so the defang is not needed for this fix.
- The stale sizing note in `bypass-dialog.ts` claiming `maybeHandleDialog()` runs only before the shell is ready; it is also called from the in-turn path at `claude-pty-shell.ts:938`. The keystroke ceiling is still correctly bounded per dialog by `BYPASS_RESET_AFTER_MISSES`, so this is a comment accuracy issue, not a behavioural one.

0xMaxMa:
Closes #436

## The bug

`ScreenModel.detectDialog()` scanned only the bottom 20 rows (`DIALOG_REGION_ROWS`) of the 50-row PTY screen. Claude Code's "Bypass Permissions mode" dialog renders at the **top** of the screen on a clean boot — there is no conversation output yet to push it down — so both detection markers fell outside the window:

```
  1  (blank)
  2  ────────────────────────────────────────────
  3    WARNING: Claude Code running in Bypass Permissions mode
  ...
 12    ❯ No, exit
 13      Yes, I accept
 14  (blank)
 15    Enter to confirm · Esc to cancel
 16-50  (blank)          <- the only rows detectDialog() looked at
```

`detectDialog()` returned `null`, `maybeHandleDialog()` returned without acting **or logging**, and the session sat at the dialog until the 120 s `STARTUP_TIMEOUT_MS` expired, shut down, and respawned into the same dialog.

Live evidence: 10 genuine startup timeouts across 3 sessions, **two of them on a build that already contained #432**.

This is **not** the bug #432 fixed. #432 corrected *which keystroke* is sent once the dialog is detected, and that fix is correct — it simply never ran here, because on a clean boot the dialog was never detected in the first place. Both are needed.

### Why it was intermittent

A boot that happens to emit some output first pushes the dialog down into rows 31-50 and detection works. A clean boot leaves it at row 3 and detection silently fails. That is the whole source of the intermittency.

## Why the fix is not "make the window bigger"

The window is the only thing standing between the auto-accept and a chat reply — or re-injected conversation history — that quotes the dialog's text. That text is real characters on a real screen, and a false accept sends `Enter`, which with the caret on its default row means **"No, exit"**: Claude Code terminates, unrecoverably.

So the window is a *proxy* for "this is the live modal, not quoted text". This PR replaces the proxy with the property it was standing in for.

## The fix

New `isBypassDialogOnScreen()` in `bypass-dialog.ts` (already dependency-free and already the owner of the option-row patterns) requires all of:

1. both option labels present as **whole rows** (not embedded in prose);
2. **exactly one** `❯` caret across those two rows;
3. the dialog's own `Enter to confirm` footer rendered **below** the option rows;
4. **nothing but blank rows below that footer**.

Rule 4 is the load-bearing one. A live modal owns the screen; quoted text always has the input box, its border and the status bar rendered underneath it — so no quoting or scrollback scenario can satisfy it, at any screen position. This is the same shape as the existing `detectRequestTooLarge()` guard, which requires the overlay's own dismiss footer precisely because the error prefix alone appears in ordinary prose.

Notably this also means `neutralizeTuiTriggers()` does **not** need widening to defang the bypass text in re-injected history: re-injected history renders inside the input box and can never satisfy rule 4.

### Defence in depth

`dialogText()` now returns the whole screen, so the "act only on what was detected" invariant can no longer be enforced by handing `decideBypassDialogAction()` a narrow region. It is enforced by applying the **identical structural test inside that function too**, which also stops the two layers drifting apart if either is edited alone. Every existing safety check there is kept — each states an invariant the decision depends on, and a keystroke sent on a wrong assumption can pick "No, exit".

### No longer silent

`maybeHandleDialog()` now logs one warning (latched, so once per run of such rounds rather than every 2 s) when the bypass markers are on screen but the dialog does not validate. The silent `null` return is why this took two investigations to find: the symptom was three layers away from the cause, with no breadcrumb in between.

## Regression test proven against the pre-fix code

The headline test is built from the **verbatim 50-row capture** above. Reverting only `dialogText()`/`detectDialog()` to their pre-fix form (leaving everything else in place, so the suite still compiles and the failures are real rather than a build error) turns **9 tests red**, including:

```
✕ detects the dialog at the TOP of an otherwise blank screen (issue #436 regression)
✕ detects the numbered rendering in both positions too (<= 2.1.247)
✕ ignores the dialog while no row is highlighted (frame still rendering)
✕ ignores a shape with a caret on both rows
✕ requires the confirm footer below the option rows
✕ ignores a footer that renders above the option rows
✕ does nothing when no row is highlighted
✕ does nothing when both rows carry a caret
✕ ignores prose that merely quotes the labels mid-sentence
```

Ten `detectDialog` tests in total now cover: top-of-screen (#436), still detected at the bottom, both Claude Code renderings in both positions, prose quoting the labels, a dialog scrolled up with live content below it, no caret, two carets, footer missing, and footer above the rows.

## Out of scope

- `DIALOG_REGION_ROWS` itself is unchanged — `parseInteractivePrompt()` still uses it for the genuinely bottom-anchored permission prompt, which is a different detector with a different liveness proof (the behavioural probe). Its doc comment now says so explicitly.
- `skipDangerousModePermissionPrompt` in the Claude Code config would stop the dialog appearing at all, but it is a machine-global file the gateway does not own, and it would mask this bug rather than fix it.

## Verification

- `npm run typecheck` — clean
- `npm test` — **4227 passed / 4227 total**, 232 suites
- Revert proof performed as described above, then restored and re-greened

### DIFF
--- a/src/shell/bypass-dialog.ts
+++ b/src/shell/bypass-dialog.ts
@@ -26,9 +26,10 @@
  * reordering must not send us to "No, exit"), and anything unparseable or
  * ambiguous produces no keystroke at all.
  *
- * This module is the pure decision — kept free of node-pty / screen imports so
- * it is cheap to unit-test in isolation, same pattern as menu-probe.ts's
- * decideProbeAttempt and menu-cancel.ts's decideMenuCancel. See
+ * This module is the pure detection + decision — kept free of node-pty / screen
+ * imports so it is cheap to unit-test in isolation, same pattern as
+ * menu-probe.ts's decideProbeAttempt and menu-cancel.ts's decideMenuCancel.
+ * ScreenModel.detectDialog() calls isBypassDialogOnScreen() below; see
  * Driver.maybeHandleDialog() in claude-pty-shell.ts for where it is wired in.
  */
 
@@ -41,6 +42,20 @@
  */
 export const BYPASS_ACCEPT_LABEL = 'Yes, I accept';
 export const BYPASS_DECLINE_LABEL = 'No, exit';
+
+/**
+ * The dialog's own confirm affordance, rendered a line or two below the option
+ * rows ("Enter to confirm · Esc to cancel"). Only the stable prefix is matched.
+ *
+ * Required by {@link isBypassDialogOnScreen} for the same reason
+ * TUI_REQUEST_TOO_LARGE_DISMISS is required by detectRequestTooLarge(): the
+ * labels alone appear in ordinary prose (this very dialog's warning text, a
+ * chat reply explaining it, re-injected history quoting it), whereas the
+ * footer is the affordance the live overlay renders — and it is exactly the
+ * affordance the accepting keystroke relies on, so gating on it keeps
+ * detection and the action consistent.
+ */
+export const BYPASS_CONFIRM_FOOTER = 'Enter to confirm';
 
 /** Arrow keystrokes used to walk the caret onto the accept row. */
 export const BYPASS_KEY_DOWN = '\x1b[B';
@@ -77,12 +92,12 @@ export const BYPASS_MAX_KEYS = 60;
  * Consecutive rounds with no dialog on screen before the keystroke ceiling is
  * considered spent on a dialog that is gone.
  *
- * Not 1: detection needs both TUI_BYPASS_PERMS markers inside the same bottom
- * region (screen.ts DIALOG_REGION_ROWS), so a repaint that shifts the box by a
- * row drops the header out of that window for a single read. Resetting on one
- * miss would zero the counter mid-dialog — precisely the case the ceiling
- * exists for. Requiring sustained absence costs the next dialog one extra
- * cooldown round at most.
+ * Not 1: detection is structural (see isBypassDialogOnScreen), so a mid-repaint
+ * frame — the option rows drawn but the footer not yet, or the caret momentarily
+ * on neither row — reads as "no dialog" for a single round while the dialog is
+ * still very much up. Resetting on one miss would zero the counter mid-dialog —
+ * precisely the case the ceiling exists for. Requiring sustained absence costs
+ * the next dialog one extra cooldown round at most.
  */
 export const BYPASS_RESET_AFTER_MISSES = 2;
 
@@ -182,6 +197,60 @@ function findRow(lines: string[], re: RegExp): OptionRow | null {
 }
 
 /**
+ * Is a live, drivable bypass-permissions dialog on this screen?
+ *
+ * This replaced a positional test — "both marker substrings inside the bottom
+ * 20 rows" — which assumed a modal is always anchored to the bottom. This
+ * dialog is not: it renders at boot, before there is any conversation to push
+ * it down, so on a clean start it sits at the TOP of the screen with the rest
+ * blank. Both markers then fall outside the window, detection returns null,
+ * nothing is pressed, and the session dies at the 120 s startup timeout and
+ * respawns into the same dialog (issue #436). Whether a given boot emitted
+ * enough output to push the dialog into the window is what made the wedge look
+ * random.
+ *
+ * The window was never really about position, though — it was a cheap proxy for
+ * "this is the live modal, not text that quotes it". Quoted text is a genuine
+ * hazard: an agent explaining this dialog, or re-injected conversation history,
+ * puts the same characters on the same screen, and a mis-aimed Enter can land on
+ * "No, exit" and kill Claude Code unrecoverably. So the proxy is replaced with
+ * the thing it was proxying for — four structural properties, all position-
+ * independent, all read off a verbatim capture of the real dialog:
+ *
+ *   1. Both option labels occupy a WHOLE row (findRow's anchored patterns), so
+ *      prose that mentions them mid-sentence never qualifies.
+ *   2. Exactly ONE of those two rows carries the ❯ caret. Zero means the frame
+ *      is still rendering (or is not a live select); two is a shape we do not
+ *      understand.
+ *   3. The dialog's own confirm footer renders BELOW the option rows.
+ *   4. Nothing but blank rows renders below that footer.
+ *
+ * (4) is the load-bearing one, and it is what the old bottom-region test was
+ * actually encoding: a live modal owns the screen. Quoted text never can — the
+ * input box, its border and the status bar always render underneath it — so
+ * scrollback, a pasted dialog, and re-injected history all fail (4) no matter
+ * where on the screen they land.
+ *
+ * Fail-safe in the same direction as before: anything unrecognised reads as "no
+ * dialog", which means no keystroke and an operator who simply sees the dialog.
+ */
+export function isBypassDialogOnScreen(screenText: string): boolean {
+  const lines = screenText.split('\n');
+  const accept = findRow(lines, ACCEPT_ROW_RE);
+  const decline = findRow(lines, DECLINE_ROW_RE);
+  if (!accept || !decline) return false;
+  // Exactly one caret: equal flags mean either none (still rendering) or both
+  // (unrecognised shape). Same rule decideBypassDialogAction() acts on.
+  if (accept.caret === decline.caret) return false;
+  const lastOptionRow = Math.max(accept.line, decline.line);
+  const footer = lines.findIndex(
+    (l, i) => i > lastOptionRow && l.includes(BYPASS_CONFIRM_FOOTER),
+  );
+  if (footer === -1) return false;
+  return lines.slice(footer + 1).every((l) => l.trim() === '');
+}
+
+/**
  * Decide the single keystroke to send at a bypass-permissions dialog, given
  * the screen region the dialog was detected in.
  *
@@ -206,6 +275,19 @@ export function decideBypassDialogAction(
     return { kind: 'wait', reason: `keystroke budget exhausted after ${state.keys} keys` };
   }
 
+  // The caller only gets here after detectDialog() fired, and both now read the
+  // whole screen — so "act only on what was detected" can no longer be enforced
+  // by handing this function a narrow region. It is enforced by applying the
+  // same structural rule here instead, which also keeps the two layers from
+  // drifting apart if either is edited alone.
+  if (!isBypassDialogOnScreen(dialogText)) {
+    return { kind: 'wait', reason: 'no live bypass dialog on screen' };
+  }
+
+  // The checks below are reached only for a screen that already satisfies that
+  // rule, so several are belt-and-braces. They are kept deliberately: each one
+  // states an invariant this decision depends on, and a keystroke sent on a
+  // wrong assumption here can pick "No, exit" and end the session for good.
   const lines = dialogText.split('\n');
   const accept = findRow(lines, ACCEPT_ROW_RE);
   if (!accept) return { kind: 'wait', reason: 'accept row not parseable on screen' };
--- a/src/shell/bypass-dialog.ts
+++ b/src/shell/bypass-dialog.ts
@@ -57,6 +57,29 @@ export const BYPASS_DECLINE_LABEL = 'No, exit';
  */
 export const BYPASS_CONFIRM_FOOTER = 'Enter to confirm';
 
+/**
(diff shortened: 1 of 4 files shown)


0xMaxMa:
## Review rounds — outcome

Two passes over 906486f's parent: a self-review, then a second **independent** adversarial review that drove the real `ScreenModel` with synthetic screens rather than trusting the comments. Every finding was verified against the code before anything was applied.

| # | Severity | Finding | Outcome |
|---|---|---|---|
| H1 | high | The two layers had **already drifted**. `detectDialog()` applied a substring gate *and* the structural predicate; `decideBypassDialogAction()` re-applied only the predicate. A screen with a perfect dialog shape but no heading returned `confirm` — a live Enter — from the layer that actually emits keystrokes. | **Fixed.** Heading check moved into `isBypassDialogOnScreen()`; `detectDialog()` is now a thin wrapper. The "cannot drift" claim is true rather than aspirational. |
| M3 | medium | No proximity requirement: the elements only had to appear *in order*, so an option row, a second option row 15 lines later and any sentence containing "Enter to confirm" below that qualified — on ordinary conversation — and yielded `confirm`. | **Fixed.** `BYPASS_MAX_ROW_GAP` / `BYPASS_MAX_FOOTER_GAP`, both read off the real capture. |
| M4 | medium | `neutralizeTuiTriggers()` was not extended to the bypass strings, so re-injected history was rejected only because Claude Code happens to paint the input box below it. | **Fixed.** Scope change vs the issue's "out of scope" note, justified: the 32MB poison loop is exactly the bug that proved a footer-requirement guard insufficient against a verbatim captured copy. |
| M2 | medium | Rule 6 ("nothing below the footer") is zero-tolerance, so a boxed rendering or a status line below the footer would not be detected — re-creating the #436 wedge. | **Accepted as accurate, deliberately not applied.** Relaxing it to tolerate border rows would admit a quoted dialog sitting above an **empty input box**, whose rows are themselves pure border characters. "No, exit" is unrecoverable; a visible dialog is not. A miss beats a false accept. Documented and pinned by a test so it is a decision on record, and the new warning makes such a miss visible instead of silent. |
| L5 | low | `bottomText()` lost its last caller. | **Fixed** — removed, with its now-disproven bottom-anchored rationale. |
| L6 | low | Stale "detection window" / "scanned region" comments. | **Fixed.** |

Also from the self-review: the new warning is gated on the same both-markers test `detectDialog()` uses rather than one marker, which appears verbatim in this repo's own `README.md:351`; and the dead `TUI_BYPASS_CONFIRM_FOOTER` re-export is dropped.

### Verified clean by the independent pass, with reasons
- **Trailing blanks.** `rowsText()` uses `translateToString(true)`, so unwritten rows are `''` and rule 6 is neither trivially true nor trivially false.
- **NBSP.** `normalize()` runs inside `rowsText()`, so a dialog padded with U+00A0 still drives correctly.
- **Detect and act cannot read different frames** — same synchronous `maybeHandleDialog()` body, and `Terminal.write()` defers parsing.
- **Mid-repaint false accept** is excluded by the existing `quietMs() >= 500` gate.
- **README** is still accurate.

### Gates
- Revert proof **re-run on the settled tree** (the code moved): **10 tests red** on the pre-fix detector, including `detects the dialog at the TOP of an otherwise blank screen (issue #436 regression)`.
- `npm run typecheck` clean.
- `npm test` — **4231 passed / 4231 total**, 232 suites.

PR #443, opened 2026-09-03
[Fix] Container app-agents silently lose Claude auth when the host settings.json is atomically replaced

### LINKED ISSUE #442: [Bug] Container app-agents silently lose Claude auth when the host settings.json is atomically replaced
## Summary

Container app-agents silently lose Claude authentication — every turn answers the end user `Not logged in · Please run /login` — because `injectAgentService()` bind-mounts host **files** into the agent container. A Docker file bind mount pins an **inode**, not a path, so when Claude Code atomically replaces `~/.claude/settings.json` on the host (write temp + rename) the container keeps reading the old, now-unlinked inode. Claude Code silently ignores a `settings.json` whose inode is unlinked, so the agent process stays healthy and keeps serving real users a login error until someone manually restarts the container.

### Root cause

`src/apps/agent-manager.ts:157-165` mounts two individual host files:

```ts
volumes: [
  `${claudeBin}:${claudeBin}:ro`,
  `${nodeBin}:/usr/bin/node:ro`,
  `${npmRoot}:${npmRoot}:ro`,
  `${homeDir}/.claude.json:${claudeJsonSeed}:ro`,                          // :161
  `${homeDir}/.claude/settings.json:${homeDir}/.claude/settings.json:ro`,  // :162
  `${workspaceDir}:/workspace`,
  `${agentMediaSource}:${agentMediaDir}:rw`,
],
```

A **file** bind mount resolves its source once, at container start, and holds that inode for the container's lifetime. A **directory** bind mount resolves entries per access. Claude Code rewrites `~/.claude/settings.json` by atomic rename whenever settings change (editing settings, installing a plugin, rotating a token), which allocates a *new* inode on the host and unlinks the old one. The container is then reading a deleted inode (`nlink=0`) — still readable, byte-identical to what it was, but stale.

Claude Code **silently ignores** a `settings.json` whose inode is unlinked. There is no warning, no crash, and nothing in the gateway logs pointing at the mount — the only visible effect is that every turn returns `Not logged in · Please run /login`.

### Live evidence

Observed on a production app-agent (a live chat bot) whose container had been running since before the host settings rewrite. Last good reply was ~3 h before the host `settings.json` was rewritten; every turn after it failed, with `"error":"authentication_failed"` / `terminal_reason: api_error` in the session log.

On this host, Claude auth comes solely from `~/.claude/settings.json` → `env.CLAUDE_CODE_OAUTH_TOKEN` (plus a custom `ANTHROPIC_BASE_URL`) — there is no `.credentials.json`, no keyring, and no auth env var on the gateway process. Control: pointing `CLAUDE_CONFIG_DIR` at an empty dir reproduces `Not logged in`; pointing it at a copy of the file succeeds.

| | inode | nlink | size | mtime |
|---|---|---|---|---|
| host `~/.claude/settings.json` | 767466 | 1 | 483 | after the rewrite |
| container's mounted view | 767440 | **0** | 438 | before the rewrite |

Isolation, run inside the *same* container with the *same* env against **byte-identical** files (same md5) — the only variable is whether the inode is still linked:

| `settings.json` fed to `claude` | nlink | result |
|---|---|---|
| fresh `cp`, via `CLAUDE_CONFIG_DIR` | 1 | **OK** |
| symlink to the bind-mounted file | **0** | **`Not logged in`** |
| `docker exec -e CLAUDE_CODE_OAUTH_TOKEN` (no settings file involved) | n/a | **OK** |

Reproduced from scratch on a throwaway `debian:12-slim` container with a single file bind mount: atomic replace on the host → the container still reads the old content at `nlink=0`; after `docker restart` → new inode, new content.

Ruled out as causes: expired token (token hash identical on host and in container), network (identical HTTP 429 from the provider for both host and container, probed with the bind-mounted `node`), config-dir resolution (the container writes its session registry into `$HOME/.claude/sessions/`, proving it resolves that directory), competing settings sources (`/etc/claude-code/managed-settings.json`, `settings.local.json` and project settings are all absent in the container), and permissions (readable by the container uid).

### Second instance of the same class

`src/apps/agent-manager.ts:161` mounts `~/.claude.json` to a read-only *seed* path. That seed mount is stale on the same live container right now — container: inode 272384, `nlink=0`, 9,346 bytes; host: inode 272375, `nlink=1`, 27,646 bytes, ~36 h newer.

The seed indirection itself is correct and must be kept: the comment at `src/apps/agent-manager.ts:107-117` explains that Claude Code atomically rewrites `~/.claude.json` *inside* the container, and a mountpoint cannot be replaced by rename (`Device or resource busy`), so the file is mounted to a seed path and copied into a writable `~/.claude.json` at container start (`:151`).

What is wrong is the claim attached to it. The comment at `:148-151` says:

> `cp` overwrites on every (re)start so the container always picks up the current host auth

That is **false** once the host file has been replaced by rename: the `cp` faithfully copies a stale seed, so even a container restart does not pick up current host state — only a `docker compose up`/recreate that re-resolves the mount does.

### Blast radius

Every container app-agent, on every gateway host, silently, whenever the host `~/.claude/settings.json` is rewritten. The failure is invisible to operators (healthy process, no logs) and fully visible to end users (every reply is a login error). Recovery today requires a manual `docker restart <agent-container>`.

Diagnostic shortcut:

```bash
docker exec <container> stat -c '%h %i' "$HOME/.claude/settings.json"
# nlink (%h) of 0  =>  the mount has gone stale
```

## Proposed Approach

Two layers — fix where the stale value is *created*, and defend where it is *consumed*, so already-broken containers recover on their own.

**1. Boundary — `src/session/process.ts` (recovers running containers with no restart)**

Read the host `~/.claude/settings.json` `env` block fresh at each spawn — a sibling to the existing `readUserScopedMcp()` at `src/session/process.ts:445-453`, which already establishes exactly this pattern — and forward the Claude auth variables into the container alongside the existing `containerEnv` / `dockerEnvFlags` at `src/session/process.ts:777-793`.

Pass them as **bare names** (`docker exec -e CLAUDE_CODE_OAUTH_TOKEN`, no `=value`), which makes Docker inherit the value from the `docker` CLI process's own environment. The value is set on the spawned process's `env` (`src/session/process.ts:809-819`) and therefore never appears on argv or in `ps`. Verified that bare-name inheritance works for `docker exec`, not just `docker run`.

Because this reads the host file by **path** on every spawn, it is immune to the inode problem by construction and needs no container restart.

**2. Source — `src/apps/agent-manager.ts` (removes the class for future containers)**

Stop bind-mounting individual host files that Claude Code atomically replaces. Mount a gateway-managed seed **directory** read-only instead, and copy out of it at container start as today. Directory bind mounts resolve per access, so a host-side rename is picked up without recreating the container.

The seed directory is populated in `injectAgentService()`, which already runs immediately before `docker compose up -d` at both call sites (`src/apps/installer.ts:1794` install, `:1992` update) — so the existing `:148-151` guarantee ("a (re)start picks up current host auth") becomes true instead of false. Update that comment to describe what the code actually does.

**3. Regression test** — `tests/unit/apps/agent-manager.test.ts` (existing `describe('injectAgentService()')` block at `:81-86`), asserting the generated compose contains no single-file mount of a Claude config file. Must be verified to fail against the pre-fix code.

## Acceptance Criteria

- [ ] A container app-agent keeps working after the host `~/.claude/settings.json` is atomically replaced, **without** a container restart or recreate.
- [ ] An already-broken container (holding an unlinked `settings.json` inode) recovers on its next turn with no manual intervention.
- [ ] The OAuth token never appears on a command line (`docker exec` argv) or in `ps` output.
- [ ] `injectAgentService()` generates no bind mount of an individual host file that Claude Code rewrites atomically (`settings.json`, `.claude.json`).
- [ ] `~/.claude.json` is still seeded via a copy into a writable path — the `Device or resource busy` constraint documented at `agent-manager.ts:107-117` is preserved, not regressed.
- [ ] The comment at `agent-manager.ts:148-151` matches actual behavior.
- [ ] A regression test covers the mount shape and is proven to fail on the pre-fix code.
- [ ] `npm run typecheck` clean, full `npm test` green.

## Out of Scope

- The `claude` and `node` binary mounts (`agent-manager.ts:158-159`) carry the same latent hazard — a `claude` self-update replaces the binary by rename, so a long-lived container could keep executing a deleted binary. Verified **not** currently stale (host and container agree on the inode), so this is latent rather than an active bug. Worth a follow-up issue.
- Any change to how the host itself obtains or stores Claude credentials.

0xMaxMa:
## Summary

A Docker **file** bind mount pins an inode, not a path. Claude Code rewrites `~/.claude/settings.json` by atomic rename, so the first host settings change left every long-lived app-agent container reading the old, now-unlinked inode. Claude Code **silently ignores** a `settings.json` whose inode is unlinked — so the container stayed healthy, logged nothing useful, and answered every real end user `Not logged in · Please run /login` until someone recreated it by hand.

This fixes it in two layers: at the **boundary**, so containers that are already broken recover on their next turn with no restart, and at the **source**, so the whole class stops being possible for future containers.

## Related Issue

Closes #442

## Changes Made

- `src/session/process.ts`: added `CONTAINER_AUTH_ENV_KEYS` and `resolveContainerAuthEnv()`, a sibling of the existing `readUserScopedMcp()`. It resolves the Claude credentials **by path, once per spawn**, and forwards them into the container. Because it resolves by path on every turn, it is immune to the inode problem by construction.
  - Credentials are passed as **bare `-e KEY` names** (no `=value`), so docker reads each value from the gateway's own environment. The token never lands on the docker argv, where any local user could read it out of `ps`.
  - `settings.json` **wins over** the gateway's own environment, which is only a fallback. This matches Claude Code, which applies the file's `env` block over whatever it inherited, and keeps the live file the single source of truth — otherwise a long-lived gateway started with an exported token would pin containers to a credential that can never be rotated without restarting the service, reintroducing the very staleness class being fixed here.
  - `settings.json` is treated as untrusted input: only clean, non-empty strings are forwarded (a NUL byte throws at spawn; a newline is never legitimate in a credential or a base URL).

- `src/apps/agent-manager.ts`: `injectAgentService()` no longer bind-mounts individual host config files.
  - `~/.claude.json` is staged into a gateway-owned seed **directory** (mode `0700`, staged file `0600`) mounted read-only. Directory bind mounts resolve entries per access, so a host-side rename is picked up by the next (re)start without recreating the container.
  - The seed indirection itself is deliberately preserved: Claude Code rewrites `~/.claude.json` *inside* the container, and a mountpoint cannot be replaced via rename (`Device or resource busy`), so the container still needs a writable copy rather than a mount.
  - The container-start copy is now `;`-separated rather than `&&`, so a failed copy surfaces on stderr (visible in `docker logs`) instead of preventing the container from starting. The copy is omitted entirely when the host has no `~/.claude.json`.
  - The `~/.claude/settings.json` mount is **dropped**. Credentials now arrive as env; the file's remaining keys (plugins, marketplaces, statusLine) are host-scoped and point at paths that were never mounted into the container anyway.
  - The comment claiming `cp` on every (re)start "always picks up the current host auth" was factually false once the host file had been replaced by rename. It now describes what the code actually does.

## Testing

- `npm run typecheck`: pass
- `npm test`: 236 suites / 4308 tests pass, zero failures

### Regression tests — proven to fail on the pre-fix code

Verified by reverting only the two source files and re-running; 7 specs go red:

`tests/unit/apps/agent-manager.test.ts`
- `never bind-mounts an individual host Claude config file (regression: stale inode => "Not logged in")` — the primary guard
- `stages the Claude config seed as a private directory the gateway owns`
- `seeds a writable ~/.claude.json via copy instead of a read-only mount at its real path`
- `omits the seed copy when the host has no ~/.claude.json`
- `mounts binaries, auth files, and workspace as volumes` (updated for the new shape)

`tests/unit/session-process.test.ts`
- `U-SP-AUTH1`: an app-agent forwards the host `settings.json` credentials into the container
- `U-SP-AUTH2`: the credential value never appears on the docker argv — and *is* present in the spawn env, so the assertion cannot pass vacuously on the pre-fix code
- `U-SP-AUTH5`: `settings.json` wins over the gateway env, which is only a fallback

`U-SP-AUTH3` (a malformed credential is not forwarded) and `U-SP-AUTH4` (a host session gets no container credential flags) are negative guards and correctly hold both before and after the fix.

## Docs

No change needed — this touches no HTTP API surface. The one `~/.claude/settings.json` reference in `API.md` (models-catalog token resolution) is host-side and untouched.

## Out of Scope

The `claude` and `node` binary mounts carry the same latent hazard — a `claude` self-update replaces the binary by rename, so a long-lived container could keep executing a deleted binary. Verified **not** currently stale, so it is latent rather than an active bug; worth a follow-up issue.

### DIFF
--- a/src/agent/model-catalog.ts
+++ b/src/agent/model-catalog.ts
@@ -16,9 +16,7 @@
  * `mcp/`. Keep the two in step.
  */
 
-import fs from 'fs';
-import os from 'os';
-import path from 'path';
+import { claudeSettingsEnv } from '../config/claude-settings';
 import type { ModelConfig } from '../types';
 
 /** A stale catalog beats a hung picker — the fetch sits in front of a UI action. */
@@ -87,12 +85,10 @@ let settingsEnvFetchedAt = 0;
 
 function settingsEnv(key: string, now: number = Date.now()): string {
   if (settingsEnvCache === undefined || now - settingsEnvFetchedAt >= CATALOG_TTL_MS) {
-    try {
-      const dir = process.env.CLAUDE_CONFIG_DIR || path.join(os.homedir(), '.claude');
-      settingsEnvCache = JSON.parse(fs.readFileSync(path.join(dir, 'settings.json'), 'utf8'))?.env ?? null;
-    } catch {
-      settingsEnvCache = null;
-    }
+    // Locating and parsing the file is shared with the other readers of it (see
+    // config/claude-settings), so CLAUDE_CONFIG_DIR is honoured identically
+    // everywhere. Only the TTL cache below is specific to the catalog.
+    settingsEnvCache = claudeSettingsEnv();
     settingsEnvFetchedAt = now;
   }
   const v = settingsEnvCache?.[key];
--- a/src/apps/agent-manager.ts
+++ b/src/apps/agent-manager.ts
@@ -107,14 +107,41 @@ export class AgentManager {
     // Claude Code rewrites ~/.claude.json atomically at startup (write temp +
     // rename over the target). Bind-mounting the host file at its real path makes
     // that rename fail with `Device or resource busy` (a mountpoint cannot be
-    // replaced via rename, regardless of ro/rw), so the session cannot persist the
-    // auth state carried by settings.json's env block and falls back to
-    // "Not logged in". Instead we bind-mount the host file to a read-only *seed*
-    // path (host file never mutated) and copy it into a writable ~/.claude.json at
-    // container start (see `command` below) so the atomic rewrite succeeds. The
-    // copy targets the literal homeDir: container-start HOME is `/`, but the
+    // replaced via rename, regardless of ro/rw), so the container needs a writable
+    // *copy* rather than a mount: we stage a gateway-owned seed and copy out of it
+    // at container start (see `command` below).
+    //
+    // The seed is a DIRECTORY, never the host file itself. A Docker *file* bind
+    // mount pins an inode rather than a path, so an atomic rewrite on the host
+    // allocates a new inode and leaves the container reading the old, unlinked one
+    // for the rest of its life — the bug that made container agents answer
+    // "Not logged in · Please run /login" after any host settings change.
+    // Directory mounts resolve entries per access, so re-staging the copy here is
+    // enough for the next (re)start to pick it up; no container recreate needed.
+    //
+    // The copy targets the literal homeDir: container-start HOME is `/`, but the
     // gateway runs turns with `docker exec -e HOME=<homeDir>` (see process.ts).
-    const claudeJsonSeed = `${homeDir}/.claude.json.seed`;
+    const containerSeedDir = `${homeDir}/.claude-seed`;
+    const seedDir = path.join(this.agentsDir, agentName, '.claude-seed');
+    fs.mkdirSync(seedDir, { recursive: true });
+    fs.chmodSync(seedDir, 0o700); // mkdir mode is umask-dependent; this is not
+    const hostClaudeJson = path.join(homeDir, '.claude.json');
+    const seededClaudeJson = fs.existsSync(hostClaudeJson);
+    if (seededClaudeJson) {
+      const dest = path.join(seedDir, '.claude.json');
+      fs.copyFileSync(hostClaudeJson, dest);
+      fs.chmodSync(dest, 0o600);
+    }
+    const seedSource = fs.realpathSync(seedDir);
+
+    // ~/.claude/settings.json is deliberately NOT mounted. On the host it exists
+    // to carry credentials, and mounting that file is exactly what broke container
+    // agents (stale inode after an atomic host rewrite). Credentials now reach the
+    // container as `docker exec -e` env resolved per turn from the live host file
+    // (see resolveContainerAuthEnv in session/process.ts), which also lets an
+    // already-broken container recover without being recreated. The file's
+    // remaining keys — plugins, marketplaces, statusLine — are host-scoped and
+    // point at paths that were never mounted into the container anyway.
 
     // Mount the agent's media directory into the container at the SAME absolute
     // path as the host. The gateway hands the agent uploaded-image paths (and the
@@ -142,13 +169,20 @@ export class AgentManager {
       `    && chown -R ${uid}:${uid} ${homeDir}`,
     ].join('\n') + '\n');
 
+    // Seed a writable ~/.claude.json from the read-only seed mount, then idle.
+    // The copy re-runs on every (re)start, and because the seed is a directory
+    // mount it resolves the *current* staged file rather than a pinned inode.
+    // Steps are `;`-separated, not `&&`: a failed copy must surface on stderr
+    // (visible in `docker logs`) without preventing the container from starting.
+    // `exec` keeps the container's PID 1 as sleep.
+    const seedCopy = seededClaudeJson
+      ? `cp ${containerSeedDir}/.claude.json ${homeDir}/.claude.json; `
+      : '';
+
     const agentService = {
       build: { context: entry.installPath, dockerfile: 'Dockerfile.agent' },
       user: `${uid}:${uid}`,
-      // Seed a writable ~/.claude.json from the read-only seed mount, then idle.
-      // `cp` overwrites on every (re)start so the container always picks up the
-      // current host auth; `exec` keeps the container's PID 1 as sleep.
-      command: `sh -c 'cp ${claudeJsonSeed} ${homeDir}/.claude.json && exec sleep infinity'`,
+      command: `sh -c '${seedCopy}exec sleep infinity'`,
       container_name: `${entry.name}-agent`,
       restart: 'unless-stopped',
       cap_drop: ['ALL'],
@@ -158,8 +192,7 @@ export class AgentManager {
         `${claudeBin}:${claudeBin}:ro`,
         `${nodeBin}:/usr/bin/node:ro`,
         `${npmRoot}:${npmRoot}:ro`,
-        `${homeDir}/.claude.json:${claudeJsonSeed}:ro`,
-        `${homeDir}/.claude/settings.json:${homeDir}/.claude/settings.json:ro`,
+        `${seedSource}:${containerSeedDir}:ro`,
         `${workspaceDir}:/workspace`,
         `${agentMediaSource}:${agentMediaDir}:rw`,
       ],
--- a/src/apps/agent-manager.ts
+++ b/src/apps/agent-manager.ts
@@ -261,8 +261,9 @@ export class AgentManager {
 
   /**
    * Remove by agent name — used in install rollback where AppEntry may not be in scope.
-   * Only removes the workspace symlink; preserves sessions/ and other data so that
-   * a reinstall picks up the same conversation history.
+   * Removes the workspace symlink and the re-stageable Claude config seed;
+   * preserves sessions/ and other data so that a reinstall picks up the same
+   * conversation history.
    */
   async deleteAgentByName(agentName: string): Promise<void> {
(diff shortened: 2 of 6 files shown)


0xMaxMa:
## Independent review — outcome

A second, independent review pass was run against this branch. Eight findings came back. Two were rated HIGH and, if correct, would have invalidated the premise of the PR; both were tested directly rather than accepted, and both are refuted by experiment. The remaining six were real and are now fixed in `f2d2067`.

### The two HIGH findings, and why they do not hold

**HIGH-1 — "Layer 1 is inert on an already-broken container; the recovery claim is false."**
**HIGH-2 — "The stated root-cause mechanism is wrong: the container reads a stale file, it does not fail to read one."**

Both rest on the same premise: that a `settings.json` `env` block always wins over the inherited process environment. That premise is correct *for a linked file*, and it was verified independently — with a live file present, a token in the file was used and a token in the environment was ignored, confirmed by two local listeners where only the file-token listener was ever hit.

The premise stops holding at exactly the state this PR is about: an **unlinked** inode. Four steps, in a throwaway container mirroring the production mount shape:

| # | State under test | Result |
|---|---|---|
| 1 | Host, isolated `HOME`, both sources set | File wins over env — reviewer's precedence premise confirmed |
| 2 | Container, mounted file at `nlink=1` | File honoured, request authenticated |
| 3 | Atomic rename on the host, content held **byte-identical** (same md5 on both sides), container inode now `nlink=0` | **`Not logged in · Please run /login` reproduced, zero requests sent** |
| 4 | Same unlinked inode, credential forwarded via `docker exec -e` | Authenticated on the env-sourced token, no "Not logged in" |

Step 3 isolates the variable completely: the file's *content* is identical, only its link count differs. If the container were reading a stale file it would have authenticated with the same bytes. It sent nothing at all — the settings file on an unlinked inode is **ignored**, not read stale. That refutes HIGH-2, and it is why step 4 works: with no file-sourced value in play, the forwarded environment is what remains. That refutes HIGH-1 — a container already in the broken state recovers on its next session spawn, with no recreate.

The review reached the opposite conclusion by testing precedence against a **freshly written** settings file, i.e. the one state in which this fix is not needed.

### Findings that were real — all fixed

- **MEDIUM-1** — `TELEGRAM_BOT_TOKEN` was on the docker argv. It is a bearer token for the agent's entire bot account, and `/proc/<pid>/cmdline` is world-readable while `/proc/<pid>/environ` is owner-only. Now forwarded by name, like the Claude credentials.
- **MEDIUM-2** — the resolver dropped malformed values silently, against the repo's own "no silent failures" rule, and produced precisely this PR's symptom with nothing in the log. Unusable values are now reported with origin and reason; a total failure to resolve warns explicitly. Key names and reasons only — never values. An unset key stays silent.
- **MEDIUM-3** — the staged config was only refreshed on install/update, so a host re-login under a different account could stay invisible to containers indefinitely. `stageClaudeSeed()` is extracted and now also called from `upsertAgent()`, which `reconcileAgents()` runs for every running app-agent at gateway start. Checked against a live container first: its seed mount was already stale (`nlink=0`, size drifted from the host file), confirming the gap was real and not theoretical.
- **LOW-1** — the doc comment presented the mechanism as inferred; it now records it as verified, and corrects the cadence (one `docker exec` serves a whole session, so the resolve is per spawn, not per turn).
- **LOW-2** — the comment's enumeration of host-scoped `settings.json` keys was incomplete and one entry was misattributed.
- **LOW-3/LOW-4/LOW-5** — vacuous test anchors, an unused export, and unquoted shell paths in the seed copy (a home directory with a space would have split into extra `cp` operands).

### Gates

- Typecheck clean.
- Full suite: **236 suites / 4313 tests passed**.
- Every fix revert-proofed individually: undoing it alone fails its guarding test.

No live container was modified and the gateway was not restarted during any of this; all experiments ran in a throwaway container that has since been removed.

PR #447, opened 2026-09-03
fix(apps): expose an in-flight boot restore so `running` no longer hides a rebuild

### LINKED ISSUE #446: [Bug] GET /api/v1/apps reports `running` for an app the boot restore is still rebuilding — in-flight restore is invisible to API consumers
## Summary

`GET /api/v1/apps` reports `running` for an app whose containers do not exist yet, because the boot-time restore deliberately suppresses reconciliation while it rebuilds — so an API consumer cannot tell "being rebuilt right now" apart from "up and serving", and has nothing to show the user during a cold rebuild that can last minutes.

## Investigation

**The suppression is intentional and correct — the gap is that it is invisible.**

1. `restoreRunningApps()` picks exactly the apps stored as `running`:

   `src/apps/installer.ts:1483`
   ```ts
   const running = apps.filter((e) => e.status === 'running');
   ```

2. …and marks that whole batch in flight before any worker starts (`src/apps/installer.ts:1494`), releasing each name in a per-app `finally` (`:1513`) plus a batch-level safety net (`:1523`):
   ```ts
   for (const entry of running) this.restoringNames.add(entry.name);
   ```

3. While a name is in `restoringNames`, `queryRuntimeStatus()` returns the **stored** status instead of querying Docker — `src/apps/installer.ts:1435-1439`:
   ```ts
   // So does a boot-time restore: its `compose build`/`up` has not created the
   // containers yet, so a read landing mid-restore sees an empty list, maps it
   // to `stopped`, and reconcileStatus() persists that underneath the restore.
   // On a cold host the window is minutes, not seconds (issue #425).
   if (this.restoringNames.has(entry.name)) return entry.status;
   ```
   That guard exists for a good reason (#425): without it a read landing mid-restore would persist `running` → `stopped`, which then disqualifies the app from every future boot restore.

4. Because the set restored is *precisely* the set stored as `running` (step 1), the status returned during the whole restore window is always the literal string `running` — for an app whose containers have not been created yet.

5. `'building'` is never involved. It is written in exactly one place, inside `runInstall` (the interactive install pipeline) — `src/apps/installer.ts:1786`:
   ```ts
   status: 'building',
   ```
   The restore path never writes it, and `queryRuntimeStatus` treats `building` as "an install owns this status" (`installer.ts:1434`), which is a different owner from a restore.

6. Restore **failure** is already surfaced to the API, via decorate-on-read in the router — `src/api/apps-router.ts:17-24`:
   ```ts
   function withRestoreFailure(
     installer: AppInstaller,
     entry: AppEntry,
   ): AppEntry & { restoreError?: string; restoreFailedAt?: string } {
     const failure = installer.getRestoreFailure(entry.name);
     if (!failure) return entry;
     return { ...entry, restoreError: failure.error, restoreFailedAt: failure.at };
   }
   ```
   applied on both read routes (`apps-router.ts:117` and `:224`), backed by `getRestoreFailure()` (`installer.ts:1535`). So the *failed* case is observable; the *in-flight* case is not. `restoringNames` (`installer.ts:402`) has no accessor at all.

**Live confirmation:** a client polling `GET /api/v1/apps` during a cold rebuild receives `"status": "running"` for every app being rebuilt, and no field anywhere in the payload distinguishes it from a healthy one. API.md documents the suppression itself (`API.md:3232`, "Boot restore") but exposes only `restoreError` / `restoreFailedAt` (`API.md:3236-3237`) — nothing for the in-flight case.

**Consequence:** every API consumer that lists apps after a gateway/host restart shows a green "running" app that is not serving yet (the route is wired but the upstream port is dead — `ECONNREFUSED`), with no way to show progress, and no way to justify a faster poll while the rebuild is in flight.

## Proposed Approach

Expose the in-flight restore the same **decorate-on-read** way the failure already is — additive, no change to `status` semantics, no new persisted state (the flag is process-local and must not reach `apps.json`).

1. **`src/apps/installer.ts`** — add an accessor beside `getRestoreFailure()` (`:1535`), reading the existing `restoringNames` set (`:402`) so there is no second source of truth:
   ```ts
   /** Whether this process's boot restore is currently bringing `appName` up. */
   isRestoring(appName: string): boolean {
     return this.restoringNames.has(appName);
   }
   ```

2. **`src/api/apps-router.ts:17-24`** — extend the existing `withRestoreFailure` decorator (rename to `withRestoreState`) rather than adding a second one, so both read routes keep going through **one** function and cannot drift:
   ```ts
   function withRestoreState(
     installer: AppInstaller,
     entry: AppEntry,
   ): AppEntry & { restoring?: true; restoreError?: string; restoreFailedAt?: string } {
     const decorated = installer.isRestoring(entry.name) ? { ...entry, restoring: true as const } : entry;
     const failure = installer.getRestoreFailure(entry.name);
     if (!failure) return decorated;
     return { ...decorated, restoreError: failure.error, restoreFailedAt: failure.at };
   }
   ```
   Both call sites (`apps-router.ts:117` for `GET /v1/apps`, `:224` for `GET /v1/apps/:name`) switch to it.

3. **Field shape** — `restoring: true` is **absent** when no restore is in flight (never `false`), matching how `restoreError` / `restoreFailedAt` already behave, so no consumer sees a new always-present key.

4. **`API.md`** — document `restoring` in the existing "Boot restore" block under `GET /api/v1/apps` (`API.md:3232-3239`), beside the two failure fields.

## Acceptance Criteria

- [ ] `AppInstaller` exposes an accessor that reports whether a named app's boot restore is in flight, reading the existing `restoringNames` set — no second source of truth, no new field on `AppEntry`, nothing persisted to `apps.json`.
- [ ] `GET /api/v1/apps` returns `restoring: true` on an entry whose boot restore is in flight, while its `status` stays whatever is stored (`running`) — the #425 guard is unchanged.
- [ ] The field is **absent** (not `false`) for every app whose restore is not in flight.
- [ ] `GET /api/v1/apps/:name` returns the same field, through the **same** decorator function as the list route (one definition, two call sites).
- [ ] An app mid-**install** still reports `status: "building"` and is not marked `restoring` — install and restore stay distinguishable.
- [ ] An app whose restore has **failed** still reports `restoreError` / `restoreFailedAt` and is no longer marked `restoring` (the in-flight marker is released in the per-app `finally` at `installer.ts:1513`).
- [ ] A test proves the transition end to end: with the name present in `restoringNames` the response carries `restoring: true` alongside `status: "running"`, and once the restore completes the flag disappears while `status` is reconciled normally. The test must fail against the current code.
- [ ] `API.md`'s "Boot restore" block under `GET /api/v1/apps` documents `restoring` (in-memory, current process only), beside `restoreError` / `restoreFailedAt`.

## Out of Scope

- Progress detail — percentage, current compose step, or streamed build logs. This issue exposes a boolean only.
- Any change to the `status` union (`running | stopped | error | building`) or to the reconciliation suppression itself — the #425 guard stays exactly as it is.
- Push delivery (SSE / WebSocket) of restore state; consumers keep polling.
- Marking apps in flight for the interactive install path — that is already covered by `status: "building"`.

## Dependencies

- Blocked by: none
- Related to: #425 (the boot-restore guard whose side effect this makes observable)

0xMaxMa:
## Summary

`GET /api/v1/apps` reported `running` for an app whose containers did not exist yet, so an API consumer could not tell a boot-time rebuild from an app that was actually serving — and had nothing to show a user during a cold rebuild that can last minutes.

The suppression behind it is intentional and stays exactly as it is: while a restore owns an app, `queryRuntimeStatus` returns the **stored** status rather than querying Docker (`src/apps/installer.ts:1439`), because a read landing mid-restore would otherwise persist `running` → `stopped` and disqualify the app from every future boot restore (#425). The problem is that the set restored is *precisely* the set stored as `running` (`installer.ts:1483`), so the status reported throughout the window is always the literal string `running`. `'building'` never applies — it is written only by `runInstall` (`installer.ts:1786`), a different owner.

Restore **failure** was already exposed by decorating the entry on read. This does the same for the in-flight case.

Closes #446

## Changes

| File | What |
|---|---|
| `src/apps/installer.ts` | New `isRestoring(name)` accessor beside `getRestoreFailure()`, reading the existing `restoringNames` set — one source of truth, so the flag cannot disagree with the suppression it describes. Nothing new is persisted to `apps.json`. |
| `src/api/apps-router.ts` | `withRestoreFailure` → `withRestoreState`, now attaching `restoring: true` alongside the failure fields. Both read routes (`GET /v1/apps`, `GET /v1/apps/:name`) keep going through the **one** function, so the list and single-app views cannot drift. |
| `tests/unit/api/apps-router.test.ts` | 5 cases under `boot restore state (#446)`. |
| `API.md` | `restoring` documented in the existing "Boot restore" block under `GET /api/v1/apps`. |

## Contract

- `restoring: true` while this process's boot restore is bringing the app up; **absent** (never `false`) otherwise — the same shape `restoreError` / `restoreFailedAt` already use, so no existing consumer gains an always-present key.
- `status` semantics unchanged, and the #425 guard is untouched. An app mid-**install** still reports `building` and is not marked `restoring`.
- `restoring` and `restoreError` are mutually exclusive: the in-flight marker is released in `restoreRunningApps()`'s per-app `finally` (`installer.ts:1513`) before a failure becomes visible.
- In-memory and process-local, like the failure fields.

## How to test

```bash
npm run build
npx jest tests/unit/api/apps-router.test.ts
```

The tests drive a **real** restore — `restoreRunningApps()` with `compose up` parked on a promise — and read both routes while it is in flight, rather than reaching into the private set:

- marks an in-flight restore on `GET /api/v1/apps` without altering its status
- omits the field entirely once the restore has finished (and `'restoring' in entry === false`)
- marks it on `GET /api/v1/apps/:name` too, through the same decorator
- leaves an app mid-install reporting `building`, not `restoring`
- reports a failed restore as `restoreError` with no lingering `restoring` flag

**Proven to fail on the pre-fix code.** With the `restoring` decoration reverted and everything else unchanged, the two behavioural cases fail (`Expected: true, Received: undefined`) and the three contract/guard cases still pass — 2 failed, 61 passed. With the fix: 63 passed.

## Gates

- `npm run build` (tsc) — clean
- `npx jest --testPathPattern=unit` — **210 suites / 4052 tests passed**

### DIFF
--- a/API.md
+++ b/API.md
@@ -3237,7 +3237,7 @@ curl -H "X-Api-Key: my-key" http://localhost:10850/api/v1/apps | jq
 > | `restoreError` | Why this process's boot restore of the app failed (compose error or timeout) |
 > | `restoreFailedAt` | ISO timestamp of that failure |
 >
-> All three are absent unless they apply (never `false`), and are in-memory only — they describe the current gateway process. `restoring` clears the moment that app's own restore ends, so it never overlaps with `restoreError`. The failure fields clear as soon as the app is observed `running`, or on an explicit start/stop.
+> All three are absent unless they apply (never `false`), and are in-memory only — they describe the current gateway process. `restoring` clears the moment that app's own restore ends, so a boot's restore surfaces as either in-flight or failed rather than both at once. The failure fields clear as soon as the app is observed `running`, or on an explicit start/stop.
 
 **`source` values:** `registry` | `custom` | `local`
 
--- a/API.md
+++ b/API.md
@@ -3237,7 +3237,7 @@ curl -H "X-Api-Key: my-key" http://localhost:10850/api/v1/apps | jq
 > | `restoreError` | Why this process's boot restore of the app failed (compose error or timeout) |
 > | `restoreFailedAt` | ISO timestamp of that failure |
 >
-> All three are absent unless they apply (never `false`), and are in-memory only — they describe the current gateway process. `restoring` clears the moment that app's own restore ends, so a boot's restore surfaces as either in-flight or failed rather than both at once. The failure fields clear as soon as the app is observed `running`, or on an explicit start/stop.
+> All three are absent unless they apply (never `false`), and are in-memory only — they describe the current gateway process. The batch to restore is marked before the gateway accepts its first request, so there is no startup window in which an app awaiting restore reports a bare `running`. `restoring` clears the moment that app's own restore ends, so a boot's restore surfaces as either in-flight or failed rather than both at once. The failure fields clear as soon as the app is observed `running`, or on an explicit start/stop.
 
 **`source` values:** `registry` | `custom` | `local`
 
--- a/API.md
+++ b/API.md
@@ -3229,14 +3229,15 @@ curl -H "X-Api-Key: my-key" http://localhost:10850/api/v1/apps | jq
 
 > **Live status:** `GET /api/v1/apps` and `GET /api/v1/apps/:name` reconcile the stored status against the live Docker runtime (`docker compose ps`) on read, so a container that crashed, was OOM-killed, was stopped from outside the gateway, or is stuck in a crash-restart loop reports `stopped`/`error` rather than a stale `running`. A `running` container (or one doing a clean/transient restart) → `running`; a container stuck `restarting` after a non-zero exit (crash-loop), an exit with a non-signal non-zero code, or a `dead` container → `error`; no containers, a clean exit, or a container force-killed by an explicit stop (exit 137/SIGKILL or 143/SIGTERM) → `stopped`. If Docker cannot be queried (daemon down, compose file missing) the last stored status is returned unchanged, and an app mid-install (`building`) is not reconciled.
 
-> **Boot restore:** an app whose containers the boot-time restore is still bringing up is **not** reconciled either — its stored status is returned as-is, so a read landing mid-restore cannot see the not-yet-created containers and write `stopped` underneath it. If that restore **failed**, the app reports `error` together with two extra fields, and the stored `running` intent is deliberately left alone so the next boot retries it:
+> **Boot restore:** an app whose containers the boot-time restore is still bringing up is **not** reconciled either — its stored status is returned as-is, so a read landing mid-restore cannot see the not-yet-created containers and write `stopped` underneath it. Because the apps restored are exactly those stored as `running`, that status alone cannot tell a rebuild in progress from an app that is actually serving, so the entry carries `restoring` while the restore is in flight. If that restore **failed**, the app reports `error` together with two further fields, and the stored `running` intent is deliberately left alone so the next boot retries it:
 >
 > | Field | Description |
 > |-------|-------------|
+> | `restoring` | `true` while this process's boot restore is still bringing the app up — its containers may not exist yet, so the reported `status` is the stored intent, not observed state |
 > | `restoreError` | Why this process's boot restore of the app failed (compose error or timeout) |
 > | `restoreFailedAt` | ISO timestamp of that failure |
 >
-> Both are absent unless a restore failed, and are in-memory only — they describe the current gateway process. They clear as soon as the app is observed `running`, or on an explicit start/stop.
+> All three are absent unless they apply (never `false`), and are in-memory only — they describe the current gateway process. `restoring` clears the moment that app's own restore ends, so it never overlaps with `restoreError`. The failure fields clear as soon as the app is observed `running`, or on an explicit start/stop.
 
 **`source` values:** `registry` | `custom` | `local`
 
--- a/src/api/apps-router.ts
+++ b/src/api/apps-router.ts
@@ -22,7 +22,18 @@ type AuthedRequest = Request & { apiKey: ApiKey };
  *
  * Both fields are absent on a healthy app rather than present-and-false, so the
  * response shape is unchanged for existing consumers. One function, used by
- * both read routes, so the list and the single-app view cannot drift apart.
+ * both read routes, so the list and the single-app view cannot drift apart —
+ * and those two are the only routes that return an AppEntry at all
+ * (`POST /:name/start|stop|restart` answers `{ name, action }`), so there is no
+ * third surface where the state could go missing.
+ *
+ * The flag is read AFTER the caller has reconciled the status, so a restore
+ * that finishes during that await is reported as the status it left behind
+ * (`running`, suppressed) with no `restoring` — stale by one tick, in the
+ * direction that understates work in progress for an app that just came up
+ * successfully. Reading it first would invert the skew and claim an app is
+ * still rebuilding after it is up; neither can be made atomic across two
+ * awaits, and this direction never reports a rebuilding app as ready.
  */
 function withRestoreState(
   installer: AppInstaller,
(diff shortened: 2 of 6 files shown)


0xMaxMa:
🔍 Code Review — #447 fix(apps): expose an in-flight boot restore (self-review)

📊 Score: 9.0 / 10.0
- Code Quality:         9.0
- Correctness:          9.0
- Production Readiness: 9.0

💪 Strengths
- `isRestoring()` reads the **existing** `restoringNames` set (`installer.ts:402`) rather than tracking a parallel copy, so the reported flag cannot disagree with the suppression at `installer.ts:1439` that it describes. Nothing new is persisted to `apps.json`.
- Extending `withRestoreFailure` into `withRestoreState` instead of adding a second decorator keeps both read routes (`apps-router.ts:128`, `:246`) on **one** function — the list view and the single-app view cannot report differently.
- `restoring` is absent rather than `false` when it does not apply, matching the `restoreError` / `restoreFailedAt` contract already in place, so no existing consumer gains an always-present key.
- The regression test drives a **real** `restoreRunningApps()` with `compose up` parked on a promise and reads the HTTP routes mid-flight, rather than reaching into the private set — it would still catch a regression that kept the accessor but dropped the decoration.

♻️ Reuse & Duplication (1 — fixed in review)
1. 🟡 `tests/unit/api/apps-router.test.ts` — the new helper hand-rolled a second `new AppInstaller(...)` purely to inject the async spawn seam, duplicating `makeInstaller()`'s construction including its sync-spawn default: two places to update the day the constructor changes shape.
   → **Fixed** in `de03a8a`: `makeInstaller()` gained an optional `asyncSpawn` param; omitting it passes `undefined`, which is the constructor's own default, so every existing caller is byte-identical in behaviour. The helper is now a 4-line wrapper.

🔴 Critical (0)
(none)

🟡 Medium (0)
(none)

🟢 Minor (2 — 1 fixed, 1 accepted)
1. `API.md` — the note claimed `restoring` "never overlaps with `restoreError`". What the per-app `finally` (`installer.ts:1513`) actually guarantees is that they do not overlap **within a boot's restore pass**; the absolute form would be wrong if `restoreRunningApps()` were ever called a second time in one process.
   → **Fixed** in `de03a8a`: reworded to the claim the code actually backs.
2. The field is a boolean — a client can say "rebuilding" but not "rebuilding for 6 minutes", because no start timestamp is exposed.
   → **Accepted.** Progress detail is explicitly Out of Scope in #446, and a timestamp is a wider contract decision (`restoreStartedAt` would want to survive the flag clearing to be useful). Better as its own issue than smuggled in here.

✅ AC Coverage (#446)
- [x] Accessor backed by the existing `restoringNames` set, no second source of truth, nothing persisted — `installer.ts:1539-1556`
- [x] `GET /api/v1/apps` returns `restoring: true` in flight with `status` unchanged — test 1
- [x] Absent (not `false`) when not restoring — test 2 asserts `'restoring' in entry === false`
- [x] `GET /api/v1/apps/:name` via the **same** decorator — test 3
- [x] Mid-install still reports `building`, not `restoring` — test 4
- [x] Failed restore reports `restoreError` with no lingering flag — test 5
- [x] Test proven to fail on the pre-fix code — 2 failed / 61 passed with the decoration reverted; 63 passed with it
- [x] `API.md` "Boot restore" block documents `restoring`

💡 Notes
- The MCP `list_apps` and `app_status` tools serialise the route's JSON straight through (`mcp/tools/apps/module.ts:248-261`), so they inherit `restoring` with no change needed.
- The #425 guard, the `status` union, and the reconciliation path are untouched — this is purely additive on the read path (one O(1) `Set.has` per entry).

Gates: `tsc` clean · `jest --testPathPattern=unit` → 210 suites / 4052 tests passed.

🏁 Verdict: ✅ Ready to merge

PR #456, opened 2026-09-04
feat: native MCP connectors — built-in + custom, per-agent enablement, managed & generic OAuth

0xthoth:
## Summary

Adds a first-class "connectors" feature to the gateway: a catalog + secret store +
config injector that lets each agent enable a set of MCP servers, with three ways
a connector gets its credentials:

1. **Managed OAuth** — a control-plane the deployer runs owns the OAuth
   client_secret and token refresh loop *off* the gateway (the gateway runs
   inside the end user's own VM/box, reachable by that user's own shell/SSH —
   a shared client_secret can't live there safely), and
   pushes a fresh `access_token` to the gateway via `POST /v1/connectors/:id/oauth/receive`.
2. **Generic custom OAuth** — for any MCP server that only offers OAuth 2.1 sign-in
   (PKCE + Dynamic Client Registration, no static API key), the gateway runs the
   *entire* flow itself, on the VM — discovery, DCR, PKCE, token exchange, and a
   background refresh sweep. No third party ever sees the token.
3. **Plain paste-token** — the simplest case: the user pastes a static API key
   directly into a custom connector's config.

Connectors resolve into each session's `mcp-config.json` fresh at spawn
(`resolveEnabledConnectors`), respecting per-agent opt-out enablement — a
connector globally connected is available to every agent by default unless that
agent explicitly disables it.

## Why

Today, wiring an MCP server into the gateway means hand-editing `config.json` and
restarting. This gives every agent a live "Connectors" surface (add, connect,
disconnect, per-agent toggle) without ever putting a vendor's client_secret on a
box the end user can shell into.

## Design notes / tradeoffs

- Secrets live in a dedicated `mcp-token.env` (0600, separate from `config.json`),
  namespaced `CUSTOM__<id>__<name>`. `config.json`'s `gateway.customConnectors`
  only ever holds the non-secret wiring (label, url, header template with a
  `{placeholder}`, which secrets it needs).
- A custom connector's pasted config is **admin-trusted, not code-reviewed** — the
  admin API key is the whole trust boundary here, same as any other admin-gated
  mutation this gateway already exposes.
- Disconnecting a connector is a *soft* disconnect when there's a secret to clear
  and something worth preserving (keeps the pasted config so reconnecting doesn't
  mean retyping it) — but a hard delete when there's nothing to preserve (a
  "No auth" connector with zero secrets, or a managed entry whose definition lives
  entirely in the external control plane anyway).
- The OAuth refresh sweep runs every 60s (same interval as the existing
  `cliPairingStore.prune()`), with backoff after a failed refresh and a hard
  give-up (clearing the connector back to "disconnected") after 3 consecutive
  failures — so a revoked refresh_token doesn't retry forever.

## Backward compatibility

Verified against a clean rebase onto current `main`: every new field on shared
types (`CustomConnectorEntry`, `AgentConfig.connectors`,
`gateway.customConnectors`, `gateway.oauthReturnUrl`) is optional and defaults
safely on a pre-existing `config.json` that has none of them. No new npm
dependency was added — everything is built on Node's own `crypto`/`fetch`/`fs`.
Nothing in this PR runs at boot beyond the (already-existing) 60s interval, which
is a no-op when `customConnectors` is empty.

## How to test

```
npx jest tests/unit/connectors.test.ts tests/unit/custom-connectors.test.ts \
  tests/unit/oauth-connectors.test.ts tests/unit/oauth-connectors-router.test.ts \
  tests/unit/mcp-oauth.test.ts tests/unit/pending-oauth-store.test.ts \
  tests/unit/oauth-refresh-sweep.test.ts tests/unit/session-process.test.ts \
  tests/unit/config-watcher.test.ts
```

Live-verified end to end against a running gateway container, including: adding a
custom connector (both plain-token and `oauth: true` shapes), the full PKCE+DCR
sign-in round trip against a real third-party MCP server, disconnect/reconnect,
per-agent enablement, and the managed push-receive path.

## Checklist

- [x] Reviewed my own code — this branch went through a dedicated internal audit
      pass after the initial implementation (three independent review passes
      covering CRUD/storage, the OAuth/PKCE/DCR flow, and gateway wiring/session
      integration), which surfaced and fixed 8 real issues before this PR: a dead
      config-write path with its own uncoordinated file lock, a way to overwrite a
      managed connector's real OAuth token through the wrong route, disconnect not
      actually disabling an OAuth connector (the refresh sweep would silently
      resurrect it), the refresh sweep never restarting sessions despite its own
      doc comment claiming it did, no backoff on repeated refresh failure, a race
      between the sweep and a manual reconnect, no per-connector error isolation
      in resolution, and unbounded DCR client registration on every Connect retry.
- [x] Tests pass
- [x] No console errors

### DIFF
--- a/e2e-connectors.cjs
+++ b/e2e-connectors.cjs
@@ -1,0 +1,82 @@
+/* Real e2e for the connectors feature — uses the COMPILED dist modules, a real
+ * HTTP server on a test port, real files in a temp HOME. No systemd, no mocks. */
+const express = require('express');
+const fs = require('fs');
+const os = require('os');
+const path = require('path');
+
+const tmp = fs.mkdtempSync(path.join(os.tmpdir(), 'cgw-e2e-'));
+const tokenEnv = path.join(tmp, 'mcp-token.env');
+const configPath = path.join(tmp, 'config.json');
+process.env.GATEWAY_MCP_TOKEN_ENV = tokenEnv;
+fs.writeFileSync(configPath, JSON.stringify({ gateway: { logDir: '/tmp', timezone: 'UTC' }, agents: [] }, null, 2));
+
+const { createConnectorsRouter } = require('./dist/api/connectors-router');
+const { resolveEnabledConnectors } = require('./dist/connectors/resolve');
+
+const ADMIN = 'admin-key-e2e';
+const apiKeys = [{ key: ADMIN, agents: '*', admin: true }];
+
+const app = express();
+app.use(express.json());
+app.use('/api', createConnectorsRouter(apiKeys, configPath));
+const server = app.listen(19099);
+
+const TOKEN = process.env.E2E_GH_TOKEN || 'ghp_dummy_e2e_test_token';
+const base = 'http://127.0.0.1:19099/api/v1';
+const H = { 'Content-Type': 'application/json', 'X-Api-Key': ADMIN };
+const log = (...a) => console.log(...a);
+
+(async () => {
+  try {
+    log('\n=== 1. GET /connectors (catalog, before connect) ===');
+    let r = await fetch(`${base}/connectors`, { headers: H });
+    log('status', r.status, JSON.stringify(await r.json()));
+
+    log('\n=== 2. auth: missing key rejected ===');
+    r = await fetch(`${base}/connectors`);
+    log('no key ->', r.status);
+
+    log('\n=== 3. POST /connectors/github/connect (real token write) ===');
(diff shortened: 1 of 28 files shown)


0xthoth:
Closing to rename the source branch — this had `-poc` in the name from early development, but it's since been through a full internal audit (3 independent review passes) with 8 fixes and full test coverage, so the name no longer fit. Reopening as a fresh PR from the renamed branch, same content.

PR #458, opened 2026-09-04
feat: native MCP connectors — per-agent enablement, managed & generic OAuth

0xthoth:
## Summary

Adds a first-class "connectors" feature to the gateway: a secret store + config
injector that lets each agent enable a set of MCP servers, with three ways
a connector gets its credentials:

1. **Managed OAuth** — a control-plane the deployer runs owns the OAuth
   client_secret and token refresh loop *off* the gateway (the gateway runs
   inside the end user's own VM/box, reachable by that user's own shell/SSH —
   a shared client_secret can't live there safely), and
   pushes a fresh `access_token` to the gateway via `POST /v1/connectors/:id/oauth/receive`.
2. **Generic custom OAuth** — for any MCP server that only offers OAuth 2.1 sign-in
   (PKCE + Dynamic Client Registration, no static API key), the gateway runs the
   *entire* flow itself, on the VM — discovery, DCR, PKCE, token exchange, and a
   background refresh sweep. No third party ever sees the token.
3. **Plain paste-token** — the simplest case: the user pastes a static API key
   directly into a custom connector's config.

Every connector is defined in `gateway.customConnectors`; there is no separate
built-in catalog. An entry pushed by an external control plane is the same shape,
distinguished only by `credentialOwner: 'external'` — the single axis that records
who is responsible for renewing the credential (`none` / `static` / `gateway` /
`external`), which is also what the UI badges on and what every status, refresh
and disconnect path branches on.

Connectors resolve into each session's `mcp-config.json` fresh at spawn
(`resolveEnabledConnectors`), respecting per-agent opt-out enablement — a
connector globally connected is available to every agent by default unless that
agent explicitly disables it.

## Why

Today, wiring an MCP server into the gateway means hand-editing `config.json` and
restarting. This gives every agent a live "Connectors" surface (add, connect,
disconnect, per-agent toggle) without ever putting a vendor's client_secret on a
box the end user can shell into.

## Design notes / tradeoffs

- Secrets live in a dedicated `mcp-token.env` (0600, separate from `config.json`),
  namespaced `CUSTOM__<id>__<name>`. `config.json`'s `gateway.customConnectors`
  only ever holds the non-secret wiring (label, url, header template with a
  `{placeholder}`, which secrets it needs). The sweep's own bookkeeping (refresh
  token, expiry, failure counters) shares the file under a separate
  `CUSTOMINT__` prefix that a pasted `{placeholder}` cannot address.
- A custom connector's pasted config is **admin-trusted, not code-reviewed** — the
  admin API key is the whole trust boundary here, same as any other admin-gated
  mutation this gateway already exposes.
- Disconnecting a connector is a *soft* disconnect when there's a secret to clear
  and something worth preserving (keeps the pasted config so reconnecting doesn't
  mean retyping it) — but a hard delete when there's nothing to preserve (a
  "No auth" connector with zero secrets, or an `external` entry whose definition
  lives entirely in the control plane that pushed it anyway).
- The OAuth refresh sweep runs on the existing 60s tick (the same interval as
  `cliPairingStore.prune()` — no new timer). Failures are split by what they mean:
  an explicit refusal from the authorization server (RFC 6749 §5.2
  `invalid_grant`, `invalid_client`, …) is **permanent** and counts toward a hard
  give-up that clears the connector back to "disconnected" after 3 in a row, so a
  revoked refresh_token doesn't retry forever. Anything else — DNS, a timeout, a
  502 — says nothing about whether the grant is still valid, so it never
  accumulates toward deletion; it backs off exponentially (5m → 6h cap) and
  recovers on its own when the provider returns. Both streaks are reported by
  `GET /v1/connectors`.

## Backward compatibility

Verified against a clean rebase onto current `main`: every new field on shared
types (`CustomConnectorEntry`, `AgentConfig.connectors`,
`gateway.customConnectors`, `gateway.oauthReturnUrl`) is optional and defaults
safely on a pre-existing `config.json` that has none of them. No new npm
dependency was added — everything is built on Node's own `crypto`/`fetch`/`fs`.
Nothing in this PR runs at boot beyond the (already-existing) 60s interval, which
is a no-op when `customConnectors` is empty.

## How to test

```
npx jest tests/unit/connectors.test.ts tests/unit/custom-connectors.test.ts \
  tests/unit/oauth-connectors.test.ts tests/unit/oauth-connectors-router.test.ts \
  tests/unit/mcp-oauth.test.ts tests/unit/pending-oauth-store.test.ts \
  tests/unit/oauth-refresh-sweep.test.ts tests/unit/session-process.test.ts \
  tests/unit/config-watcher.test.ts
```

Live-verified end to end against a running gateway container, including: adding a
custom connector (both plain-token and `oauth: true` shapes), the full PKCE+DCR
sign-in round trip against a real third-party MCP server, disconnect/reconnect,
per-agent enablement, and the external push-receive path.

## Checklist

- [x] Reviewed my own code — this branch went through a dedicated internal audit
      pass after the initial implementation (three independent review passes
      covering CRUD/storage, the OAuth/PKCE/DCR flow, and gateway wiring/session
      integration), which surfaced and fixed 8 real issues before this PR: a dead
      config-write path with its own uncoordinated file lock, a way to overwrite a
      gateway-owned connector's real OAuth token through the wrong route, disconnect not
      actually disabling an OAuth connector (the refresh sweep would silently
      resurrect it), the refresh sweep never restarting sessions despite its own
      doc comment claiming it did, no backoff on repeated refresh failure, a race
      between the sweep and a manual reconnect, no per-connector error isolation
      in resolution, and unbounded DCR client registration on every Connect retry.
      Subsequent review rounds on the PR itself added: a single shared
      `config.json` write lock replacing four private ones, the transient/permanent
      failure split above, a DCR registration cache that no longer poisons the
      client_id the sweep refreshes with, soft vs strict `mcp-token.env` reads, and
      removal of the unused built-in catalog tier.
- [x] Tests pass
- [x] No console errors


0xthoth:
Supersedes #456 (same content — closed only to rename the source branch away from its old `-poc` name).

Issue #450, opened 2026-09-03
[Bug] systemd install/update gaps: on-failure exit-0 blocks auto-restart, and a conflicting unit at another scope goes undetected


0xMaxMa:
## Summary
Two related systemd-integration gaps that both surfaced from the same real incident: (A) an API-triggered update kills the gateway via a graceful `exit 0` into a `Restart=on-failure` unit, so it never comes back on its own; (B) `service install` has no awareness of a same-named unit at a different systemd scope, so it can silently create a second, conflicting, also-enabled unit.

## Investigation

**Part A — update-triggered exit doesn't survive `Restart=on-failure`**
Live journal evidence: gateway died at 12:41:59 UTC after an API-triggered update, no restart attempt logged, manual start at 12:55:52. `src/api/packages.ts:174` sets `warning: 'service will restart'` then `src/api/packages.ts:177` does `setTimeout(() => process.kill(process.pid, 'SIGTERM'), 500)`. `src/shutdown-signals.ts:71` handles that signal via the clean-shutdown path, calling `exit(0)`. But `src/cli/commands/service.ts:149` (`renderSystemdUnit()`) — the template `service install` itself writes — emits `Restart=on-failure`, and systemd's `on-failure` policy does not restart on a graceful `exit 0`. pm2 is unaffected: its `autorestart` restarts on any exit code, so the `isManaged` assumption in `packages.ts` only holds for pm2.

**Part B — `service install` doesn't detect a conflicting unit at another scope**
`systemdInstall()` (`src/cli/commands/service.ts:257-296`) only ever inspects/writes the user-scope unit path from `unitPath()` (`~/.config/systemd/user/claude-gateway.service`); `systemdState()` (`src/cli/commands/service.ts:235-248`) only queries `systemctl --user`. Neither checks system scope (`/etc/systemd/system/claude-gateway.service`). Observed in production: a system-scope unit already existed (from an externally provisioned system unit, `Restart=always`, correct `gateway start` invocation), then someone ran `claude-gateway service install`, which created a second, independent, also-enabled user-scope unit without detecting or touching the first. Both enabled = race for the port on next reboot, the loser restart-loops to `failed`. A plain warning isn't enough — a user who proceeds past it still ends up in the same broken state. `pm2Install` (`src/cli/commands/service.ts:375-411`) doesn't have this specific gap (self-heals by deleting any existing same-named PM2 entry first, line 394) but has no cross-manager awareness either — noted for context, not fixed here.

[...]

PR #459, opened 2026-09-04
Add system-scope install mode to `claude-gateway service install`

### LINKED ISSUE #457: [Feature] Add system-scope install mode to `claude-gateway service install`
## Summary
`claude-gateway service install` only supports installing a user-scope systemd unit (`~/.config/systemd/user/`), which is the right default for an interactive, no-sudo install on a personal machine — but automated/infra provisioning workflows that run as root and need the service to run under a fixed system account have no supported path through the CLI, and end up hand-rolling their own systemd unit instead of reusing `renderSystemdUnit()`.

## Investigation
`src/cli/commands/service.ts` hard-codes the user-scope architecture throughout:
- `unitPath()` (line 176) always resolves to `~/.config/systemd/user/claude-gateway.service`
- `unitFlagState()` (line 241) always scopes `systemctl` queries with `--user` when called from the install/status/uninstall paths for systemd
- `renderSystemdUnit()` (line 131) hard-codes `WantedBy=default.target` (line 153) and has no `User=` directive — the unit always runs as whichever account happens to invoke `systemctl --user`, which only works for an interactive login-scoped install
- `systemScopeConflict()` (line 277) refuses a user-scope install if a system-scope unit with the same name (`UNIT_NAME`, line 26) is already enabled/active — correct for the current user-scope-only design, but this guard would incorrectly fire against a legitimate system-scope install of the same unit if that mode existed
- `systemdInstall()` (line 295) always prints the `loginctl enable-linger` hint (line 350) — only relevant when the unit is user-scope and needs to survive logout; irrelevant (and confusing) for a system-scope unit that starts at boot via `multi-user.target` regardless of any login session
- `systemdInstall()` calls `systemctl --user enable --now`, which does not restart a unit that is already active — a repeated install used to apply an updated unit definition (rendered from new flags, or a newer CLI version) silently keeps the stale unit running until something else restarts it

There's also no way today to customize the rendered unit's `After=` ordering, add an `EnvironmentFile=`, or add extra `Environment=` lines beyond what `renderSystemdUnit()` hard-codes — any consumer needing those has to bypass the CLI's installer entirely and write the unit file itself, missing out on any future hardening `renderSystemdUnit()` picks up (e.g. the `OOMPolicy=continue` fix already in this CLI).

## Proposed Approach
1. Add a `--scope system|user` flag to `service install/status/uninstall` (default `user`, preserving current behavior exactly).
2. When `scope=system`: `unitPath()` resolves to `/etc/systemd/system/claude-gateway.service`; all `systemctl` calls drop `--user`; fail fast with a clear error if the process is not running as root (never auto-escalate via sudo).
3. Add a required-with-`scope=system` `--run-as <user>` flag. Add `User=<run-as>` to the rendered unit, and derive `WantedBy=multi-user.target` automatically for system scope (vs. today's `default.target` for user scope) instead of a separate flag.
4. Extend `renderSystemdUnit()`'s input to accept `after?: string[]` (append to `After=`, on top of the default `network-online.target`), `envFile?: string` (adds `EnvironmentFile=-<path>`), and `extraEnv?: Record<string,string>` (additional `Environment="KEY=VALUE"` lines) — exposed via repeatable `--after <target>` / `--env-file <path>` / repeatable `--env KEY=VALUE` flags.
5. Skip `systemScopeConflict()` and the `loginctl enable-linger` hint entirely when `scope=system` — a system-scope install is establishing that exact unit, not colliding with a foreign one, and linger is meaningless for a `multi-user.target` unit.
6. When the unit is already active and its rendered content differs from what's on disk, `daemon-reload` then `restart` instead of only `enable --now`, so a repeated install actually applies the update.
7. Add matching test coverage in `tests/unit/cli-service.test.ts` / `cli-service-force-flag.test.ts` for every new flag, following the existing suite's convention (structural assertions on `renderSystemdUnit()` output, confirmation-gate behavior, `--print` previews).

## Acceptance Criteria
- [ ] `--scope user` (or omitted) behaves identically to today — no regression to the existing default path
- [ ] `--scope system` writes to `/etc/systemd/system/claude-gateway.service`, refuses non-root, requires `--run-as`, sets `User=` and `WantedBy=multi-user.target`
- [ ] `--after`, `--env-file`, `--env` flags are reflected correctly in the rendered unit for both scopes
- [ ] System-scope install does not trigger `systemScopeConflict()` against itself, and does not print the linger hint
- [ ] Re-running install against an already-active unit with changed content restarts it; re-running with unchanged content does not needlessly restart
- [ ] `service status` / `service uninstall` work correctly against a system-scope unit
- [ ] All new flags have test coverage; full existing suite still passes

## Out of Scope
- Any auto-elevation (sudo prompting) from within the CLI — the caller is responsible for already running as root when requesting `--scope system`
- PM2 scope changes — this issue is systemd-only

0xMaxMa:
## Summary
Adds a `--scope system` mode to `claude-gateway service install/status/uninstall` for automated/infra provisioning that needs the gateway to run under a fixed system account, alongside the existing default user-scope install. Also adds `--after`/`--env-file`/`--env` to customize the generated unit, switches the generated unit to `Type=exec`, distinguishes a health-check timeout from an actual install failure by exit code, and fixes a latent bug where a repeated install with changed unit content silently left the stale unit running.

## Related Issue
Closes #457

## Changes Made
- `src/cli/commands/service.ts`:
  - `renderSystemdUnit()` takes a new optional `SystemdUnitOptions` (scope, runAs, after, envFile, extraEnv); defaults to the prior exact user-scope output so `--scope user` (or omitting it) is a strict no-op.
  - `--scope system` writes `/etc/systemd/system/claude-gateway.service`, sets `User=<run-as>` and `WantedBy=multi-user.target`, refuses non-root (never auto-escalates via `sudo`), and requires `--run-as <user>`.
  - `WorkingDirectory=`/`HOME=`/the config path resolve to `--run-as`'s real home (looked up via `getent passwd`, NSS-aware), not the installing root process's — including `~`-expansion in `--config`/`--env-file`, and `$GATEWAY_CONFIG` from root's own environment is not consulted for `--scope system`. A freshly-created `WorkingDirectory` is `chown`'d to `--run-as`, and ownership is reasserted if the directory already exists but belongs to a different user (e.g. `--run-as` changed between installs).
  - `--after <target,...>`, `--env-file <path>`, `--env KEY=VALUE,...` (comma-separated, since the shared CLI parser has no repeatable-flag concept) add extra `After=` targets, an `EnvironmentFile=-<path>`, and extra `Environment=` lines respectively. Validated against NUL/newline injection and (for `--after`) embedded whitespace; `--env` rejects malformed pairs and refuses to override `HOME`/`PATH`/`GATEWAY_CONFIG`.
  - The generated unit uses `Type=exec` (was `Type=simple`) — reported "started" only once the actual `execve()` succeeds, not right after `fork()`.
  - `crossScopeConflict()` guards both directions: installing user scope while an externally-provisioned system-scope unit is active (the original #450 case), and installing system scope while a prior user-scope install is still active. Neither direction fires against a repeat install of the same scope.
  - `service status`/`service uninstall` auto-detect scope when `--scope` is omitted (mirrors `--manager` auto-detection); refuse if both scopes are simultaneously installed rather than silently picking one. `install` always defaults to `user` regardless of what already exists.
  - Re-running `install` against an already-active unit whose rendered content changed now restarts it (and re-enables it, in case it had drifted to active-but-disabled) instead of leaving the stale unit running.
  - `install`'s exit code distinguishes `1` (install/enable failed, or a validation/confirmation gate refused — nothing written) from `2` (install/enable succeeded but `/health` never answered in time), for both the systemd and pm2 paths.
- `README.md`: documents `--scope system` usage, the exit-code contract, and replaces the now-obsolete "reinstall then manually restart" migration note with the automatic-restart-on-change behavior.
- `scripts/gen-cli.ts` / `CLI.md`: regenerated service-install doc block for the new flags and exit codes.
- `tests/unit/cli-service.test.ts` / `tests/unit/cli-service-force-flag.test.ts`: extensive new coverage — scope validation/root-check/run-as requirement and home resolution (including `getent` failure, `~`-expansion, chown-on-create vs. chown-reassignment), unit rendering with `--after`/`--env-file`/`--env` and injection guards, cross-scope conflict in both directions, scope auto-detection and the dual-scope-installed refusal, `Type=exec`, the exit-code-2 distinction (using fake timers to avoid a real 10s health-poll wait), and the restart-vs-enable regression. Every regression test in this PR was proven to fail on the pre-fix code (temporarily reverting the fix, confirming red, then restoring) before being accepted.

## Review history
This PR went through 6 review passes before merge: self-review, three independent `/code-review` passes, one manual `/gateway-code-review` pass, plus a round of fixes from an externally-reported review. Each pass found and fixed real issues — see the commit history for the specific findings and fixes from each round.

## Testing
- `npm run typecheck`: pass
- `npm run gen-cli:check`: up to date
- `npm test`: 237 suites / 4457 tests pass

### DIFF
--- a/CLI.md
+++ b/CLI.md
@@ -93,13 +93,23 @@ never mistaken for one.
 
 | Command | Description |
 |---------|-------------|
-| `claude-gateway service install [--manager systemd\|pm2] [--config <path>] [--yes] [--print] [--force]` | Generate and start a service |
-| `claude-gateway service status [--manager systemd\|pm2]` | Report installed/enabled/active state as JSON |
-| `claude-gateway service uninstall [--manager systemd\|pm2] [--yes]` | Stop and remove the service |
+| `claude-gateway service install [--manager systemd\|pm2] [--scope user\|system] [--run-as <user>] [--after <target,...>] [--env-file <path>] [--env KEY=VALUE,...] [--config <path>] [--yes] [--print] [--force]` | Generate and start a service |
+| `claude-gateway service status [--manager systemd\|pm2] [--scope user\|system]` | Report installed/enabled/active state as JSON |
+| `claude-gateway service uninstall [--manager systemd\|pm2] [--scope user\|system] [--yes]` | Stop and remove the service |
 
 - `systemd` (the default) installs a **user** unit at `~/.config/systemd/user/claude-gateway.service` —
   no `sudo`, and it runs as the user that owns `~/.claude-gateway`. Run
   `loginctl enable-linger <user>` once if it must survive logout.
+- `--scope system` (systemd only) installs a root-owned unit at
+  `/etc/systemd/system/claude-gateway.service` instead, for automated/infra provisioning that needs
+  a fixed system account. Requires already running as root (never escalates via `sudo`) and
+  `--run-as <user>`, which becomes the unit's `User=`; `WantedBy=` is `multi-user.target` instead of
+  `default.target`, and the `loginctl` hint is skipped. It never triggers the system-scope conflict
+  check below against itself.
+- `--after <target,...>`, `--env-file <path>`, and `--env KEY=VALUE,...` customize the generated unit
+  (either scope): extra `After=` ordering targets, an `EnvironmentFile=-<path>` for secrets that never
+  appear in the unit text, and additional non-secret `Environment=` lines. `--env` refuses to override
+  `HOME`, `PATH`, or `GATEWAY_CONFIG`.
 - Every path in the generated unit (node, entry point, config, working directory) is absolute, and
   `ExecStart` always uses the explicit `gateway start` command. No secrets are written into the
   unit — the gateway reads `~/.claude-gateway/.env` itself.
@@ -108,9 +118,12 @@ never mistaken for one.
   and refuse to run non-interactively without it.
 - `install` verifies `/health` on the **local bind address**, and `uninstall` reports the state the
   manager actually reports afterwards — never the state that was intended.
-- (systemd) `install` refuses by default if a `claude-gateway.service` unit already exists and is
-  enabled or active at *system* scope (e.g. from provisioning outside this CLI) — it prints the exact
-  `sudo systemctl disable --now claude-gateway.service` to resolve it; pass `--force` to install anyway.
+- (systemd, user-scope installs) `install` refuses by default if a `claude-gateway.service` unit
+  already exists and is enabled or active at *system* scope (e.g. from provisioning outside this
+  CLI) — it prints the exact `sudo systemctl disable --now claude-gateway.service` to resolve it;
+  pass `--force` to install anyway.
+- Re-running `install` against an already-active unit whose rendered content changed restarts it
+  automatically; unchanged content leaves the running unit alone.
 - After installing, `gateway restart`/`stop` detect and drive that same service.
 
 ## Versions & updates
--- a/CLI.md
+++ b/CLI.md
@@ -118,6 +118,11 @@ never mistaken for one.
   and refuse to run non-interactively without it.
 - `install` verifies `/health` on the **local bind address**, and `uninstall` reports the state the
   manager actually reports afterwards — never the state that was intended.
+- `install` exit codes: `0` fully healthy, `1` install/enable itself failed (or a validation/
+  confirmation gate refused — nothing was written), `2` install/enable succeeded but `/health`
+  never answered within the poll window. Distinguishing `1` from `2` by exit code alone means a
+  caller doesn't have to parse the JSON result on stdout just to tell "didn't happen" apart from
+  "happened, health unconfirmed".
 - (systemd, user-scope installs) `install` refuses by default if a `claude-gateway.service` unit
   already exists and is enabled or active at *system* scope (e.g. from provisioning outside this
   CLI) — it prints the exact `sudo systemctl disable --now claude-gateway.service` to resolve it;
--- a/README.md
+++ b/README.md
@@ -136,13 +136,38 @@ The systemd path writes `~/.config/systemd/user/claude-gateway.service`. Run
 `loginctl enable-linger $USER` once if it must keep running while you're logged out.
 The unit sets `OOMPolicy=continue` so that an OOM-killed child process (e.g. a dev server an
 agent spawned on its own) doesn't take the whole gateway down with it — only `Restart=always`
-restarting the *gateway's own* process is intended. Installs from before this was added need to
-re-run `claude-gateway service install` to regenerate the unit file *and then*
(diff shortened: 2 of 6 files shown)


0xMaxMa (commented):
🔍 **Code Review — PR #459, re-reviewed fresh against current HEAD (`23ae1a1`, 2026-09-04 10:45 UTC)**

Full fresh checkout + fetch this time (not reusing a stale one) — read the entire `service.ts` line by line, ran the actual test suite, typecheck, and doc-gen check myself rather than trusting the PR description.

📊 **Score: 9.5 / 10.0**
- Code Quality: 9.5
- Correctness: 9.5
- Production Readiness: 9.5

## 💪 Strengths
- **Every finding from the last two review rounds is fixed, each with dedicated test coverage** — not just patched inline:
  - Exit code `2` (`EXIT_HEALTH_TIMEOUT`) now distinguishes "installed fine, health unconfirmed" from `1` ("install/enable itself failed") — tested for both systemd and PM2 paths, including the negative case (still `1` when `enable` itself throws).
  - `WorkingDirectory`/`HOME`/config path now correctly resolve through `--run-as`'s real home via `getent passwd`, not the installer's — plus two more subtleties I hadn't even flagged: a `~` in `--config`/`--env-file` now expands against the *run-as* user's home (`expandHomeAs()`), and `$GATEWAY_CONFIG` from the installing (root) process's own env is correctly *not* consulted for `--scope system` (still honored for `--scope user`, tested both ways).
  - `WorkingDirectory` ownership reassignment now also handles the case where the directory already exists but is owned by a *different* prior `--run-as` user (not just "did this call just create it") — tested.
  - `Type=exec` restored with a clear rationale comment.
- **Verified myself, not just trusted the PR body:** `npx tsc --noEmit` clean, `npx jest tests/unit/cli-service*.test.ts` → 94/94 pass, `npm run gen-cli:check` → docs up to date. README/CLI.md both document the 3-way exit code contract accurately.
- Test names literally cite where feedback came from (`issue #457 review, external report`, `manual /code-review round`) — good practice for traceability, and confirms the review loop is actually closing the loop, not just accumulating comments.

## ♻️ Reuse & Duplication (0)
No duplication or over-engineering found.

## 🔴 Critical (0)
## 🟡 Medium (0)
## 🟢 Minor (0)

Nothing left to flag. Every item from my two earlier passes on this PR (WorkingDirectory/HOME, Type=exec, exit-code ambiguity) is now resolved and tested.

## 🏁 Verdict: ✅ Ready to merge (as far as an external reviewer without merge rights can call it) — no outstanding correctness issues, full test/typecheck/doc-gen coverage confirmed directly, not assumed.

PR #477, opened 2026-09-09
[Fix] Replace Stale Pending Wizard Draft on Start

### LINKED ISSUE #475: fix(api): wizard/start should replace a stale pending draft instead of 409
## Context
The getpod Create-Agent wizard (Crown-Labs/getpod#2493) locks an agent ID for ~30 minutes when the user navigates **Back → Next** or **Cancel → reopen** with the same ID: `POST /api/v1/agents/wizard/start` returns `409 "Wizard for agent 'X' is already in progress"`.

The web-side redundant-call bug is fixed in getpod. But two cases need this gateway change to fully resolve:
- Back → change the description, **same ID** → Next
- Cancel/✕ → reopen the wizard with the **same ID** → Next

(getpod#2493 lists the gateway change as outside its scope — separate repo.)

## Root cause
`src/api/router.ts`, `POST /v1/agents/wizard/start` (~L1118 on `main`):

```ts
if (wizardStore.findByAgentId(id)) {
  res.status(409).json({ error: `Wizard for agent '${id}' is already in progress` });
  return;
}
```

This rejects **any** existing draft for the ID — including a stale `pending` draft the *same* user abandoned. A `pending` draft (`wizard-state.ts`, `step: 'pending' | 'confirmed' | 'complete'`) has not been written to disk and owns no on-disk agent, so it is safe to replace. Today only `POST .../complete` deletes a draft, so an abandoned one lingers the full 30-min TTL.

## Fix
Replace a `pending` draft on start (reuse the existing `WizardStore.delete`); keep the 409 for a `confirmed`/`complete` draft, and leave the `agentConfigs.has(id)` "already exists" check untouched.

## Acceptance criteria
1. `start` for an ID whose existing draft is `pending` → **201** with a new `wizardId`; the old draft is removed.
2. `start` for an ID with a `confirmed`/`complete` draft → still **409**.
3. The `agentConfigs.has(id)` "already exists" check is unchanged.
4. Unit test covering (1) + (2).

## Notes
- `WIZARD_MAX_CONCURRENT` (=2) still caps concurrent generations, so replacing a pending draft opens no load hole.
- Implemented + tested on branch `fix/2493-wizard-pending-replace` (2 unit tests; the replacement test fails on current `main`).

yuuyui:
## Summary
Replace an abandoned `pending` Create-Agent wizard draft when `/v1/agents/wizard/start` is called again for the same agent ID. This removes the 30-minute lockout after Cancel or after changing the description during Back → Next, while preserving conflicts for drafts that have already advanced beyond the pending stage.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor / Tech debt
- [ ] Documentation

## Related Issues
Closes #475

## Changes Made
- `src/api/router.ts`: delete a matching pending wizard draft before generating its replacement; continue returning 409 for confirmed or complete drafts.
- `tests/unit/wizard-router.test.ts`: cover pending replacement, confirmed conflicts, old-draft removal, and the existing-agent guard.
- `tests/unit/api-avatar-wizard.test.ts`: align API-level expectations with the pending-replacement contract and retain the confirmed-draft 409 regression lock.

## How to Test
1. Start a wizard for a new agent ID and leave it in the pending state.
2. Start another wizard with the same agent ID.
3. Confirm the request returns 201 with a new `wizardId` and the prior pending draft is removed.
4. Advance a draft to confirmed and retry the same agent ID.
5. Confirm the request still returns 409.

Expected: abandoned pending drafts are replaced, while active confirmed/complete drafts and existing agents remain protected from duplicate starts.

### Automated verification
- `npm test -- --runInBand` — 273 suites / 5374 tests passed
- `npm run typecheck` — passed

## Checklist
- [x] Reviewed my own code
- [x] Tests pass
- [x] No new console errors

### DIFF
--- a/API.md
+++ b/API.md
@@ -737,7 +737,7 @@ curl -X POST \
 |--------|------|
 | 400 | Invalid `id` format or missing `prompt` |
 | 403 | Not an admin key |
-| 409 | Agent or wizard already exists for this ID |
+| 409 | Agent already exists, or a `confirmed`/`complete` wizard is already in progress for this ID (a `pending` draft is replaced instead) |
 | 429 | Too many wizard starts in progress (max 2 concurrent) |
 | 500 | Claude generation failed |
 
--- a/src/api/router.ts
+++ b/src/api/router.ts
@@ -1380,19 +1380,16 @@ export function createApiRouter(
       return;
     }
     const existingWizard = wizardStore.findByAgentId(id);
-    if (existingWizard) {
-      if (existingWizard.step === 'pending') {
-        // A prior draft for this id never advanced past generation — the user hit
-        // Back/Cancel then restarted with the same id (#2493). It owns no on-disk
-        // agent, so replace it rather than dead-ending the resume with a 409.
-        // WIZARD_MAX_CONCURRENT still caps generation load.
-        wizardStore.delete(existingWizard.wizardId);
-      } else {
-        // 'confirmed'/'complete': files were already written — a real conflict.
-        res.status(409).json({ error: `Wizard for agent '${id}' is already in progress` });
-        return;
-      }
+    if (existingWizard && existingWizard.step !== 'pending') {
+      // 'confirmed'/'complete': files were already written — a real conflict.
+      res.status(409).json({ error: `Wizard for agent '${id}' is already in progress` });
+      return;
     }
+    // A prior 'pending' draft for this id never advanced past generation — the
+    // user hit Back/Cancel then restarted with the same id (#2493). It owns no
+    // on-disk agent, so it's safe to replace — but don't delete it until the
+    // replacement is guaranteed below, so a 429/500 doesn't destroy it for nothing.
+    // WIZARD_MAX_CONCURRENT still caps generation load.
 
     if (wizardStartsInFlight >= WIZARD_MAX_CONCURRENT) {
       res.status(429).json({ error: 'Too many wizard starts in progress, please retry later' });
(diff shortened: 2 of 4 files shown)


yuuyui:
## 🔍 WTB Review — Self-review at `7c4cbee`

**Score: ~9.4/10 — Ready to merge**

- Critical: 0
- Medium: 0
- Minor: 1 (documented below, non-blocking)

### Summary
Reviewed `src/api/router.ts`'s wizard-start pending-draft replacement against `wizard-state.ts`'s lifecycle (`pending` → `confirmed` → `complete`). Replacing a `pending` draft is safe by construction: `pending` means generation completed but files were never written to disk (that only happens on `/confirm`), so there's no real agent to protect. `confirmed`/`complete` correctly still 409, since those own on-disk state.

Verified the error copy `Agent '${id}' already exists` against the web client's `/already exists/i` regex directly in source — they're in sync (previously this was reported as a fact from memory; confirmed against the actual code this pass).

### Tests
Both `tests/unit/wizard-router.test.ts` (new) and the updated case in `tests/unit/api-avatar-wizard.test.ts` cover the two branches correctly:
- pending draft → replaced (fresh wizardId, old one gone)
- confirmed draft → still 409, untouched

The old test that previously asserted a flat 409 for a same-id restart was correctly updated to match the new contract rather than left as dead coverage.

### Remaining Minor (non-blocking)
The pending draft is deleted *before* the `WIZARD_MAX_CONCURRENT` check. If the system is at capacity when a same-id restart comes in, the old draft is deleted and then the request still 429s — the user loses the old draft for nothing. This is an accepted trade-off, not a new gap: the web client already handles a failed replacement start by clearing its own stale `wizardId` and letting the next attempt start fresh (tested in `agent-wizard-back-next.test.tsx`), so the failure mode is "retype and regenerate," not a dead end. Worth reordering the delete to after the concurrency check if this comes up in practice, but not blocking.

### Full suite
Confirmed clean earlier this session: 273 suites / 5374 tests passed, 0 TypeScript errors. No source changes since.

Issue #472, opened 2026-09-09
[Bug] gateway.publicUrl is never provisioned, and its absence fails silently until generate_image


0xKirin:
## Summary

`gateway.publicUrl` is required for every feature that hands out a public link, but no code path in the package ever writes it, and its absence is not reported by config load, share mint, or `doctor` — the gap surfaces only as a late, unrelated-looking failure inside `generate_image`.

**Priority:** Medium

## Background / Context

On a freshly provisioned host that sits behind an external reverse proxy (the documented production topology — see `README.md:719-723`), `gateway.publicUrl` is absent from `~/.claude-gateway/config.json`. The gateway starts cleanly, `doctor` reports healthy, and the share API mints shares successfully. The only visible symptom appears at the end of an unrelated user-facing flow:

```
generate_image: image reference sharing requires gateway.publicUrl to be configured.
```

The value cannot be inferred by the process — `README.md:721` states this explicitly ("The process cannot infer its own public URL — it binds localhost by default and sits behind a reverse proxy"). On such deployments the public hostname lives in the reverse proxy's configuration, outside the host process. So the only actor that can supply it is the provisioning layer.

That provisioning gap is out of scope here (see **Out of Scope**). What is in scope, and fully inside this package, is the **silent degradation**: three layers that could have surfaced the missing value do not. That is what turns a one-line config fix into a multi-hour investigation.

Blast radius on an affected host: per `README.md:332-348` and `README.md:746`, at minimum `generate_image` reference edits, `share_file` (and its deprecated `share_image` alias), and the `/cli` terminal viewer.

**Verified against `main` @ `4df7458` (v1.8.9).**

## Reproduction

1. Install the gateway fresh on a host behind an external reverse proxy, with no `publicUrl` in config.
2. Start the gateway — starts cleanly, no warning.
3. Run `claude-gateway doctor` — reports healthy.
4. `POST /api/v1/shares` (or the `share_file` MCP tool) with a local image — succeeds, returns a `token`, but **no `url`**.
5. Call `generate_image` with a local or `artifact:` reference.
6. Fails: `generate_image: image reference sharing requires gateway.publicUrl to be configured.`

Steps 2–4 all look healthy. Only step 6 fails.

## Evidence

**Nothing writes the key.**

- `src/config/bootstrap.ts:34` — first-boot bootstrap writes only `{ logDir, timezone }`:
  ```ts
[...]

PR #478, opened 2026-09-09
fix(share): inbound-Host fallback for share URLs so i2i/i2v work without gateway.publicUrl

0xTH0R:
## Problem

Image-to-image (`generate_image` with a ref) and image-to-video (`generate_video` with a source frame) **fail closed** on any pod whose `gateway.publicUrl` is unset: the share mint returns a `token` but no `url`, and the media tools reject with *"requires gateway.publicUrl to be configured."*

This is not rare. getpod writes `gateway.publicUrl` into each pod's `config.json` **only at provision/create/reprovision** (getpod #1939, 2026-07-30). Every pod created before that date never got it, and the incus-agent is **masked at activation** for tenant isolation — so those running pods can't be reached with `incus exec` to backfill the config. Hand-editing each pod (config edit + gateway-service restart) is the only current remedy, and it doesn't scale.

#472 added a clearer *error message* for this state but never resolved the underlying failure. This PR does.

## Fix

Learn the gateway's public base from **real inbound requests** instead of requiring it in config.

Production pods sit behind Traefik, whose routing rule pins the `Host` to the pod's own FQDN — so the `Host` header of a genuine external request **is** the gateway's public host. This is the exact trust model `line-webhook-router.ts` already relies on (`.public-base`, hardcoded `https` because the `web-vm` entrypoint doesn't trust `X-Forwarded-Proto`).

- A small global middleware persists the learned host to a **gateway-level `.public-base`** file (one host per pod, shared across agents).
- The share **mint endpoint** reads it as the base for the `url` convenience field whenever `gateway.publicUrl` is unset. Configured `publicUrl` always wins.

**One change fixes `generate_image` i2i, `generate_video` i2v, and `share_file` uniformly** — no per-pod config edit, no VM/gateway restart, no domain-munging, and no dependency on a gateway update ever reaching the old pods.

## Safety

- Loopback / internal / dotless hosts (MCP subprocess at `127.0.0.1`, Docker health checks) are rejected, so they can never clobber the learned public host.
- Scheme hardcoded `https`; host validated against the same charset as the LINE path.
- Write is best-effort (never throws into the request path), atomic (temp+rename), idempotent, and memoed in-process so the steady-state hot path costs **zero** disk I/O.
- The `url` field is only a convenience string; the capability is the token, which is minted and validated exactly as before. A spoofed Host can't reach this backend because Traefik routes by Host.

The media-tool errors now tell the user to **resend** (the host is learned from an inbound message) rather than "restart the gateway."

## Tests

- `public-base` helper: host extraction/validation (accepts real FQDN incl. `X-Forwarded-Host` precedence and port; rejects loopback/internal/dotless/invalid), atomic idempotent persist, read-back, overwrite-on-change.
- Mint endpoint: `url` filled from `.public-base` when no config base; config base wins over a stale fallback.
- Updated the `#472` i2i/i2v error-message assertions.

`tsc --noEmit` clean; affected unit suites (`public-base`, `share-router`, `share-normalize`, `video-generate-deliver`, `video-module`, `image-module`, `line-image`) all green (141 tests).

Relates to #472.

### DIFF
--- a/mcp/tools/image/module.ts
+++ b/mcp/tools/image/module.ts
@@ -512,15 +512,19 @@ export class ImageModule implements ToolModule {
         await revokeSharesBestEffort(minted.map((m) => m.share_id));
         return fail('generate_image: image share service returned a mismatched share count.');
       }
-      // Provider fetches these refs over HTTPS, so each needs an absolute URL —
-      // which the share API only fills when gateway.publicUrl is configured.
+      // Provider fetches these refs over HTTPS, so each needs an absolute URL.
+      // The share API fills `url` from gateway.publicUrl, or — when that is unset
+      // — from the public host learned off a real inbound (Traefik-fronted)
+      // request. If it is still empty the gateway hasn't seen an external request
+      // yet this run, so ask the user to resend rather than push a broken URL.
       if (minted.some((m) => !m.url)) {
         await revokeSharesBestEffort(minted.map((m) => m.share_id));
         return fail(
-          'generate_image: image reference sharing requires gateway.publicUrl to be configured ' +
-          '(set it in ~/.claude-gateway/config.json to your externally reachable base URL ending ' +
-          'in /gateway, then restart the gateway). As a workaround, pass a publicly reachable ' +
-          'https:// image URL instead of a local path or artifact ref.',
+          'generate_image: could not resolve the gateway public URL for the reference image yet. ' +
+          'Send your request again (the gateway learns its public URL from an inbound message). ' +
+          'If it keeps failing, set gateway.publicUrl in ~/.claude-gateway/config.json to your ' +
+          'externally reachable base URL ending in /gateway, or pass a publicly reachable https:// ' +
+          'image URL instead of a local path or artifact ref.',
         );
       }
     }
--- a/mcp/tools/video/module.ts
+++ b/mcp/tools/video/module.ts
@@ -414,15 +414,19 @@ export class VideoModule implements ToolModule {
     }
     const item = minted[0];
     if (!item) return fail('generate_video: share service returned no share for the source frame.');
-    // The provider fetches this over HTTPS, so it needs an absolute URL — which the
-    // share API only fills when gateway.publicUrl is configured.
+    // The provider fetches this over HTTPS, so it needs an absolute URL. The
+    // share API fills `url` from gateway.publicUrl, or — when that is unset — from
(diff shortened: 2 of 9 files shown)


0xMaxMa:
Thanks for fixing the real gap here — pods that never got `gateway.publicUrl` written to their config genuinely fail closed on image-to-image/image-to-video, and deriving the base from the inbound request instead of requiring a per-pod config edit is a reasonable direction.

**Security concern: `X-Forwarded-Host` is trusted with no origin check, which is unsafe for self-hosted deployments.**

`externalHostFromHeaders()` (`src/config/public-base.ts:43-63`) reads `req.headers['x-forwarded-host']` first, falling back to `req.headers.host`. The only validation is a format regex (`HOST_RE`) and a loopback/internal-address filter — there is no allowlist of trusted proxy IPs and no equivalent of Express's `trust proxy`. The accepted value is then persisted pod-wide via `persistPublicBase()` (`public-base.ts:78`) and later used to mint share URLs.

In a deployment where the reverse proxy in front of the gateway is configured to strip/overwrite `X-Forwarded-Host` before forwarding, a client-supplied value never reaches this code. But `claude-gateway` is a general-purpose open-source project — anyone can self-host it (e.g. on their own VPS) behind a reverse proxy config that doesn't strip this header, behind a different reverse proxy entirely, or with the gateway reachable directly with no reverse proxy at all. In any of those setups, an external client can set `X-Forwarded-Host: evil.com` directly on their request, and this code will accept and persist it as the pod's public base URL. Every share link minted afterward then points at the attacker's domain — a phishing/URL-confusion vector, and it also clobbers the base URL for every agent on the box, not just the requester's.

**This shouldn't merge as-is without a safeguard.** A few options that would close the gap without giving up the fallback:
- An explicit opt-in — only trust `X-Forwarded-Host` when the operator has configured a trusted-proxy allowlist (IP or CIDR), analogous to Express's `trust proxy` or similar trusted-proxy-count/list hardening used for `X-Forwarded-For`/`X-Forwarded-Host`. Skip straight to `Host` (or refuse to learn at all) when untrusted.
- Or, at minimum, a loud warning in the README / `config.template.json` making it explicit that any reverse proxy terminating in front of the gateway **must** strip/overwrite `X-Forwarded-Host` (and ideally `Host`) before this fallback is safe to rely on — plus fail-safe-off by default rather than trust-by-default.
- Open to other mitigations you'd propose — happy to help implement whichever direction you'd rather take, including a PR against this branch if useful.

PR #482, opened 2026-09-11
feat(wechat): add WeChat channel — QR-linked personal account via Tencent's iLink Bot API

0xthoth:
## Summary

Adds a WeChat channel: link a personal WeChat account via QR code (no credential to type in, same UX shape as the existing WhatsApp/Baileys channel) and receive/reply to direct messages through it. Built against Tencent's own unofficial "iLink Bot API" — the same protocol documented in `Tencent/openclaw-weixin` and independently implemented in `NousResearch/hermes-agent`'s Weixin adapter. Confirmed working end-to-end against a real WeChat account (link, text messages both directions, inbound images and files/PDFs with AES-128-ECB decrypt).

## Scope (v1)

- **DM only, no groups** — iLink cannot deliver WeChat group events at all (confirmed in both reference projects' own docs), so there's no group policy/allowlist for this channel, unlike Telegram/Discord/LINE/Slack/WhatsApp.
- **QR-only linking** — no pairing-code (typed-code) alternative; iLink doesn't offer one.
- **Single account per agent** — no multi-account list.
- **Inbound images + files (PDF etc.)** — downloaded and AES-128-ECB decrypted, staged into the same MediaStore path every other channel already uses. Outbound media sending is out of scope for v1 (`wechat_reply` stays text-only).
- Gated behind `WECHAT_CHANNEL_ENABLED` — a hard kill switch, since this is a third-party bridge no deployer of this gateway controls (protocol changes, an undocumented verification-code state this integration can't act on yet, etc).

## Notable protocol findings (not documented anywhere else I could find)

- A successful `getupdates`/`sendmessage` response omits the `ret` field entirely — it's only present on failure. Checking `ret !== 0` (matching how it reads in the docs) treats every success as a failure.
- `redirect_host` (returned on a `scaned_but_redirect` QR-status response) is a bare hostname, not a full URL as the field name implies.
- The real CDN host actually used in production traffic is `.wechat.com` (e.g. `novac2c.cdn.wechat.com`, `ilinkai.wechat.com`), not just the `.weixin.qq.com` hosts documented in `Tencent/openclaw-weixin`'s own protocol.md and CDN allowlist.
- A decrypted **file** payload (not image) carries a few extra leading bytes ahead of the real content that neither reference project's docs mention — handled here via the sender-declared `file_item.len` to trim to the exact plaintext length.
- `getUpdates` failures against this endpoint are mostly ordinary Cloudflare edge timeouts on an idle long-poll, not real errors — this channel's retry cadence matches `Tencent/openclaw-weixin`'s own official client (flat short retry, one fixed pause after N failures, then reset) rather than exponential backoff, which otherwise starves the poll loop of attempts.

`base_info.bot_agent` is a per-deployer config field (`AgentConfig.wechat.botAgent`, defaulting to `"claude-gateway"`), mirroring `channels.openclaw-weixin.botAgent` in the reference project rather than a fixed literal — this gateway has no single downstream product identity to hardcode.

## Test plan

- [ ] `npm run build` and `npx tsc --noEmit` clean
- [ ] `npx jest tests/unit/wechat-manager.test.ts tests/unit/ilink-client.test.ts tests/unit/wechat-access.test.ts tests/unit/api-router-wechat.test.ts tests/unit/wechat-mcp.test.ts` — new/updated unit tests covering the protocol client, manager state machine, access control, REST routes, and MCP tool
- [ ] Manual: link a real WeChat account via QR, confirm a DM round-trip (send/receive text), confirm an inbound image and a PDF are both correctly decrypted and readable by the agent
- [ ] Confirm `WECHAT_CHANNEL_ENABLED=false` (or unset) disables the channel with no other behavior change


0xMaxMa:
## Review outcome (fresh pass on the current HEAD)

**Code quality: good.** SSRF protections for CDN media downloads (allowlist guard + redirect-manual/opaqueredirect check), timeouts on every external call, and a permissive PKCS#7 unpad matching the reference implementation are all in place, with test coverage for each edge case.

**Kill switch — verified fixed.** Commit `e42ab24` flips the flag from `WECHAT_CHANNEL_ENABLED` (opt-in) to `WECHAT_CHANNEL_DISABLED` (opt-out, enabled by default). Confirmed both `startLinking()` and `resumeIfLinked()` respect the new flag correctly, and the test suite (`isWeChatChannelEnabled()` assertions) was updated to match. No lingering gap here.

**Docs gap — now fixed in this commit (`251de50`).** README.md and API.md previously had zero mentions of the WeChat channel, unlike Telegram/Discord/LINE/Slack/WhatsApp which are all documented there. Added:
- A `## WeChat Channel API` section in API.md covering the `status`/`link`/`unlink`/`send`/`pending` endpoints, the `wechat` config block (`dmPolicy`/`dmAllowlist`/`pairing`/`botAgent`), and the `WECHAT_CHANNEL_DISABLED`/`ILINK_BASE_URL`/`WECHAT_ILINK_FAKE` environment variables.
- A WeChat row in README.md's Channel Conditions & Limitations table, plus a "WeChat limits" bullet block matching the existing Telegram/LINE/WhatsApp sections.

**Verification:**
- `npx tsc --noEmit` — clean
- Targeted WeChat/iLink unit tests (5 suites) — 130/130 passed

No other correctness/security/duplication/consistency issues found. This closes out the outstanding item from the prior review — nothing further to fix on my end.

0xMaxMa:
Follow-up fix: `WECHAT_CHANNEL_DISABLED` now also covers `sendMessage()` and the poll loop (`src/wechat/manager.ts`), not just `startLinking()`/`resumeIfLinked()`.

Previously, flipping the kill switch on a process that already had a linked session running had no effect until restart — `sendMessage()` kept sending and the poll loop kept dispatching inbound messages, which contradicted the "kill instantly, no redeploy needed" doc comment on the flag.

Changes:
- `sendMessage()` now checks `isWeChatChannelEnabled()` up front and throws the same error `startLinking()` does when disabled.
- `runPollLoop()`'s loop condition now rechecks `isWeChatChannelEnabled()` every iteration and exits cleanly (no crash) once it's flipped off mid-run.

Added unit tests for both cases (kill switch flipped after linking, before `sendMessage`; kill switch flipped mid-poll-loop). Full test evidence:
- `tsc --noEmit`: clean
- Targeted WeChat suites (wechat-manager, wechat-access, wechat-mcp, api-router-wechat, ilink-client): 132/132 passed
- Full unit suite: 253 suites / 5186 tests, all passed

Commit: 6b91850

PR #504, opened 2026-09-17
Fix permanent attachment failures blocking channel ingress

### LINKED ISSUE #503: [Bug] Permanent attachment failure blocks channel ingress indefinitely
## Summary
A permanently unavailable attachment blocks the durable channel ingress queue indefinitely, so subsequent messages and commands never reach the agent, including after a gateway restart.

## Investigation
Confirmed on gateway 2.0.3 in a read-only production investigation:
- A Telegram video attachment was 44,702,906 bytes. A direct `getFile` check returned HTTP 400 with `Bad Request: file is too big`.
- The initial video and 22 subsequent messages remained in the receiver spool. Later text messages and `/new` commands were not admitted to the orchestrator.
- Thousands of `ATTACHMENT_UNAVAILABLE` callback warnings occurred at approximately one-second intervals. Restarting the gateway replayed the same persisted failing item.
- The most recent decision for the affected conversation was completed, rather than a model call still running. This failure occurs before model invocation.

Relevant code (investigated on `feat/500-lean-orchestration-token-report`; verify line positions against the implementation branch):
- `src/orchestration/channel-media.ts:10-18`: Telegram `getFile` non-success responses become generic `ATTACHMENT_UNAVAILABLE`; the size check happens only after a successful response. The permanent provider rejection loses its classification.
- `src/agent/runner.ts:748-800,890-896`: channel admission/media handling throws before acceptance; the callback catch returns HTTP 503, even for permanent attachment failures. Its log misleadingly labels these as body parsing failures.
- `mcp/tools/receiver-spool.ts:13,37-52`: flush runs each second, sorts persisted messages oldest first, and breaks on any non-success response. One poison item blocks subsequent items. Persistence reproduces the blockage after restart.
- `mcp/tools/telegram/receiver-server.ts:30-32,1119`: Telegram uses this durable spool; replies can reference the same failed attachment and must also be handled.

## Proposed Approach
1. Classify permanent attachment rejection separately from transient network/provider failures. Use known size metadata for early checks, but enforce actual download limits independently. Preserve safe provider status/reason without exposing credentials or signed URLs.
2. Admit the user's text/caption and an explicit unavailable-attachment result durably when an attachment cannot be processed. Inform the user clearly once; do not pretend the file was read, silently lose the message, or invoke analysis against missing content.
3. Define callback acknowledgement/error semantics so a permanently rejected attachment cannot poison the queue. Retain durable delivery and deduplication. Handle partial albums and unavailable quoted attachments as well as direct attachments.
4. Add bounded backoff and recovery for transient failures, with suitable isolation so one failing conversation does not block unrelated conversations. Preserve ordering where needed; do not fix this by indiscriminately dropping queue entries or changing `break` to `continue` without delivery semantics.
5. Provide safe handling of existing persisted poison entries after upgrade, avoiding automatic execution of a large stale backlog without considering user intent. Improve diagnostics with safe message identifiers and retry classification.

## Acceptance Criteria
- [ ] An oversized Telegram attachment yielding `400: file is too big` cannot block later text or commands.
- [ ] Caption/text remains available and the missing attachment is represented honestly; the user receives a concise actionable notice without repeated spam.
- [ ] A reply referencing the failed file and a mixed valid/invalid album do not block ingress or duplicate successful deliveries.
- [ ] Existing persisted failing entries recover safely across restart/upgrade; no silent data loss or duplicate admission.
- [ ] Transient network errors, rate limits, and provider 5xx responses retry with bounded backoff rather than a one-second hot loop.
- [ ] One conversation's failure cannot indefinitely block another conversation in the same receiver.
- [ ] Regression tests cover permanent rejection followed by text and `/new`, transient recovery, restart, deduplication, partial attachments, and quoted attachment failures.
- [ ] Shared receiver/media paths used by other channels are checked for the same failure mode, while respecting each provider's own limits.

## Out of Scope
- Dashboard redesign and token accounting.
- Increasing Telegram's hosted Bot API download limit or silently discarding queued production messages.

0xMaxMa:
## Summary
Prevent permanent attachment failures from blocking durable channel ingress, while preserving message text, readable files, ordering, and safe recovery of existing receiver journals.

Closes #503

## Changes
- Classify media failures by provider status and error type rather than matching error wording. Known sizes and actual download streams are bounded independently; permanent failures become explicit unavailable attachments, while network/408/429/5xx failures remain retryable.
- Persist retry backoff and isolate conversation queues. Preserve album quiet windows and freeze attempted album batches so late members cannot mutate an already admitted message.
- Deduplicate provider receipts before fetching files again, persist one unavailable-attachment notice per input, and clean staging only after durable admission. Voice transcription requires the original direct file; quoted audio cannot replace unavailable new audio.
- Archive stale startup conversation queues and conflicting envelopes without executing old commands or tasks. Preserve original envelopes, including expanded legacy albums, under stable recovery identities and ask users to resend still-needed requests/files.
- Document the internal callback contract, recovery policy, and bounded album receipt retention.
- Align integration fixtures with the existing partial-rejection and best-effort voice contracts (three stale assertions also failed on unchanged main), supply the callback fixture's durable store, and make HTTP-heavy review fixtures tolerate full-suite load without losing cancellation behavior.

## Verification
- Regression proof against the pre-fix implementation: four behavioral assertions fail (permanent rejection, rate-limit response, stale recovery, conversation isolation); implementation restored afterward.
- Added real callback/SQLite restart coverage, partial and quoted media coverage, and durable receiver retry/album-race regressions. Three additional regression assertions fail on the pre-review head for quoted-voice promotion, expanded legacy album loss, and conflicting-envelope retry.
- Self-review and independent review completed; the bounded additional review verified both review fixes without new findings.
- `npm run typecheck`: pass.
- `npm test`: 393 suites / 6,450 tests passed at `33cb1f0`, before the metadata-isolation follow-up.
- Metadata-isolation follow-up: typecheck passed; 38 affected Jest tests and 1 Bun test passed. Eight new malformed-album cases failed before the fix and now cover retained bytes, conversation isolation, restart, and ordered recovery after repair.
- `bun test mcp/tools/receiver-spool.test.ts`: 1 test passed.

## Behavior notes
- Startup queues containing an entry older than five minutes are archived as a batch; fresh arrivals continue normally.
- Late album members arriving after a batch was attempted are admitted separately under their own message IDs. Acknowledged album member receipts are bounded to 24 hours/latest 1,000; pending payloads and retry state are not expired.
- No provider download limit is increased, and no queued production messages are silently discarded.


0xMaxMa:
### Code review

No unresolved findings after self-review, independent review, and a bounded additional review of the fixes.

Resolved:
1. An unavailable direct voice recording could promote readable quoted audio into a fresh voice request. Transcription now requires the original direct attachment and uses its reference explicitly. [Voice admission](https://github.com/0xMaxMa/claude-gateway/blob/33cb1f0ce5d0446be3851afb136a89b09a9ffcd9/src/agent/runner.ts#L838-L847).
2. Expanded legacy albums could lose late members when their first message ID matched an older receipt. Conflicting envelopes are now preserved through the existing non-executing recovery path with stable deduplication and one notice per original message. [Conflict recovery](https://github.com/0xMaxMa/claude-gateway/blob/33cb1f0ce5d0446be3851afb136a89b09a9ffcd9/src/agent/runner.ts#L770-L791), [legacy receipt comparison](https://github.com/0xMaxMa/claude-gateway/blob/33cb1f0ce5d0446be3851afb136a89b09a9ffcd9/src/orchestration/store.ts#L176-L190).

Verification on this exact tree: typecheck passed; 393 suites / 6,450 tests passed; the receiver Bun test passed. Four core regression assertions fail on pre-fix main, and three review regression assertions fail on the pre-review head. Implementations were restored before the final green run.

0xMaxMa:
### Follow-up review

Fixed the remaining malformed-album finding in `cce97cd887ee146f1e030c248205b5def72d91d1`. Nested album member metadata is validated inside the existing per-record error boundary; invalid records and their conversation tails remain on disk while unrelated conversations continue. Sending uses the already validated IDs instead of parsing again outside that boundary. [Implementation](https://github.com/0xMaxMa/claude-gateway/blob/cce97cd887ee146f1e030c248205b5def72d91d1/mcp/tools/receiver-spool.ts#L86-L102).

Verification: all 8 new regression cases failed before the fix; 38 affected Jest tests, 1 Bun test, and typecheck pass after the fix. The tests also verify restart behavior and ordered recovery after repairing the retained record. README remains current; the API recovery documentation now explicitly covers malformed album metadata.

Repeated review of the latest PR found no further findings. The full suite was previously green at `33cb1f0`; this narrow follow-up was verified with the affected suites.

PR #507, opened 2026-09-18
[Fix] Drop HEARTBEAT.md from API session rules

### LINKED ISSUE #506: [Bug] API system prompt embeds the literal "HEARTBEAT.md", triggering a provider 400 on every response
## Summary

`API_SOURCE_RULES` in `src/orchestration/source-policy.ts:3` embeds the literal filename `HEARTBEAT.md` into the system prompt of every API/web session. When an agent's own `AGENTS.md` also contains `HEARTBEAT_OK` — which the stock agent template does — the provider rejects the request with `400 Third-party apps now draw from your extra usage, not your plan limits.`, and every response from that agent fails.

**Priority:** Medium

## Background / Context

Every chat with an agent through the web UI or `POST /api/v1/agents/<id>/messages` returned:

```
The response could not be completed. Please check /tasks for any pending work.
```

The orchestration store only persists the code (`INFERENCE_FAILED`), never the provider text. The real provider response was recovered by replaying the exact `claude` CLI argv the gateway spawns (read from `/proc/<pid>/cmdline`) and reading the result line:

```json
{"type":"result","is_error":true,"api_error_status":400,
 "result":"API Error: 400 Third-party apps now draw from your extra usage, not your plan limits. Add more at claude.ai/settings/usage and keep going."}
```

The same account, same `ANTHROPIC_BASE_URL` and same OAuth token work fine from the normal Claude Code CLI — the CLI does not pass `--append-system-prompt`, so the trigger never reaches the provider there.

Bisecting the 102 KB system prompt the gateway injects isolated the trigger to two sentences. All results below are deterministic (3/3 repeats each), reproduce on both `claude-opus-5[1m]` and `claude-haiku-4-5`, and are independent of prompt size:

| system prompt | result |
| --- | --- |
| `Read HEARTBEAT.md if it exists. If nothing needs attention, reply HEARTBEAT_OK.` | **400** |
| `Read HEARTBEAT.md if it exists (workspace context). Follow it strictly.` alone | ok |
| `If nothing needs attention, reply HEARTBEAT_OK.` alone | ok |
| the two sentences with `HEARTBEAT` renamed to `PULSE` | ok |
| a real agent's full 102 KB `AGENTS.md` | **400** |
| the same 102 KB with only `HEARTBEAT.md` → `PULSE.md` | ok |
| 102 KB of lorem ipsum | ok |

So the provider is reacting to that content pair, not to prompt length, model, or any other spawn flag (`--tools ''`, `--strict-mcp-config`, `--mcp-config` and the inline `--settings` object all pass when tested individually).

The gateway cannot control what an operator writes in `AGENTS.md`, but it does control `API_SOURCE_RULES`. Removing the filename from that sentence removes one half of the trigger pair from every API session permanently, so an operator who writes `HEARTBEAT_OK` in their agent persona no longer ends up with a fully broken agent and an error message that does not explain why.

## Acceptance Criteria

- [ ] `API_SOURCE_RULES` in `src/orchestration/source-policy.ts` no longer contains the literal string `HEARTBEAT.md`
- [ ] The rule still forbids memory-store writes in API sessions (behaviour unchanged, wording only)
- [ ] An API session whose agent `AGENTS.md` contains `HEARTBEAT_OK` (and no `HEARTBEAT.md`) completes a normal turn instead of failing with `INFERENCE_FAILED`
- [ ] `HEARTBEAT_OK` is left untouched everywhere — `src/cron/scheduler.ts:173` matches it to suppress heartbeat replies
- [ ] A regression test asserts `API_SOURCE_RULES` does not contain `HEARTBEAT.md`

## Out of Scope

- Surfacing the provider text to the user. Already handled on `main` by #505: `src/orchestration/process-turn.ts:169` now populates `providerMessage` from `event.result` via `structuredProviderMessage()`, and `src/orchestration/inference-errors.ts:15-18` returns the sanitized text. The released `v2.0.4` build (`d981995`) predates that commit, which is why the generic message is still what users see in the field.
- Renaming `HEARTBEAT.md` as a workspace file, or changing `src/agent/workspace-loader.ts`. The file itself is fine; only the literal inside the injected system prompt matters.
- `src/agent/create-agent-prompts.ts`. Verified that wizard-generated agents do not emit a heartbeat section in `AGENTS.md` — the affected personas in this report were hand-written.

## How to Implement

`src/orchestration/source-policy.ts:3` — drop the filename from the sentence:

```diff
-export const API_SOURCE_RULES = `Memory updates are not supported in API sessions. Do not create or update MEMORY.md, USER.md, HEARTBEAT.md or other memory stores. ...
+export const API_SOURCE_RULES = `Memory updates are not supported in API sessions. Do not create or update MEMORY.md, USER.md or other memory stores. ...
```

The constant is injected at `src/orchestration/bridge.ts:171`, where it is concatenated into `sourceRules` for every session that is not memory-writable — i.e. all API/web sessions.

Add a unit assertion alongside the existing orchestration tests that `API_SOURCE_RULES` does not match `/HEARTBEAT\.md/`, so the string cannot come back silently.

## Dependencies

- Blocked by: none
- Related to: #505 (preserve provider errors across chat and API responses), #501 (actionable provider quota errors)

## Notes

- Repro harness: read `/proc/<pid>/cmdline` of the spawned `claude` process while a request is in flight (it matches `--include-partial-messages` and `--append-system-prompt`), rebuild the argv, and re-run it with a stream-json user turn. The `{"type":"result","is_error":true}` line carries the provider text that the store discards.
- `session subprocess exited` with `code: 143` appears for every agent on every turn and is normal SIGTERM — it is not a fault signal and cost some time to rule out during this investigation.
- Workaround for operators on a released build: remove `HEARTBEAT.md` from the agent's `AGENTS.md` (keeping `HEARTBEAT_OK` is fine). Verified across 8 agents — all returned HTTP 200 afterwards.

0xKirin:
## Summary

`API_SOURCE_RULES` named `HEARTBEAT.md`, so the literal filename reached the system prompt of every API/web session through `bridge.ts`. When an agent's own `AGENTS.md` also instructed it to reply `HEARTBEAT_OK`, the provider rejected the request with `400 Third-party apps now draw from your extra usage, not your plan limits.` and every response from that agent failed. Removing the filename keeps the memory-store prohibition intact while taking one half of the trigger pair out of every API session.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor / Tech debt
- [ ] Documentation

## Related Issues
Closes #506

## Changes Made
- `src/orchestration/source-policy.ts`: drop `HEARTBEAT.md` from the `API_SOURCE_RULES` sentence — it now reads "Do not create or update MEMORY.md, USER.md or other memory stores." Wording only; the prohibition itself is unchanged.
- `tests/unit/orchestration/persona-worker.test.ts`: regression test asserting `API_SOURCE_RULES` does not match `/HEARTBEAT\.md/` while still carrying the memory-store prohibition, `MEMORY.md` and `USER.md`.

## Evidence

The trigger was isolated by bisecting the 102 KB system prompt the gateway injects. Every row below is deterministic (3/3 repeats), reproduces on both `claude-opus-5[1m]` and `claude-haiku-4-5`, and is independent of prompt size:

| system prompt | result |
| --- | --- |
| `Read HEARTBEAT.md if it exists. If nothing needs attention, reply HEARTBEAT_OK.` | **400** |
| `Read HEARTBEAT.md if it exists (workspace context). Follow it strictly.` alone | ok |
| `If nothing needs attention, reply HEARTBEAT_OK.` alone | ok |
| the two sentences with `HEARTBEAT` renamed to `PULSE` | ok |
| a real agent's full 102 KB `AGENTS.md` | **400** |
| the same 102 KB with only `HEARTBEAT.md` → `PULSE.md` | ok |
| 102 KB of lorem ipsum | ok |

The provider text was recovered by replaying the exact `claude` CLI argv the gateway spawns (read from `/proc/<pid>/cmdline`) and reading the `{"type":"result","is_error":true,"api_error_status":400}` line — the orchestration store persists only the `INFERENCE_FAILED` code.

## How to Test

1. Check out this branch.
2. `npx jest tests/unit/orchestration/persona-worker.test.ts`

Expected: 3 tests pass, including `API source rules forbid memory writes without naming the heartbeat file`.

To reproduce the original failure end to end, give an agent an `AGENTS.md` containing `HEARTBEAT_OK` and send it a message through `POST /api/v1/agents/<id>/messages`. On a build without this change the response fails; with it, the session completes normally.

## Notes

- `HEARTBEAT_OK` is deliberately untouched — `src/cron/scheduler.ts:173` matches it to suppress heartbeat replies, so renaming it would break heartbeat suppression.
- `src/orchestration/inference-errors.ts` is out of scope: #505 already routes the provider text to the user via `providerMessage`, so once that ships the underlying 400 will at least be legible. This PR removes the cause rather than the symptom.
- `src/agent/create-agent-prompts.ts` and `src/agent/workspace-loader.ts` are untouched — wizard-generated agents do not emit a heartbeat section, and the workspace file itself is not the problem.

## Checklist
- [x] Reviewed my own code
- [x] Tests pass (`npx jest tests/unit/orchestration/persona-worker.test.ts` — 3/3)
- [x] No console errors


0xKirin:
## 🔍 Code Review — self-review

**📊 Score: 8.7 / 10.0**

- Code Quality: 9.0
- Correctness: 8.5
- Production Readiness: 8.5

### 💪 Strengths

- Fixes the cause rather than the symptom — it removes one half of the trigger pair from a string the gateway controls, instead of chasing every operator's `AGENTS.md`.
- The test matches the surrounding file style (compact, no space after `,`) and the comment explains *why* rather than *what*, so a future reader is less likely to delete the assertion as pointless.
- The test pins both directions: the filename is gone **and** the original prohibition (`MEMORY.md`, `USER.md`, the leading sentence) survives — so "fixing" it by deleting the whole sentence would fail.
- Minimal blast radius: one string, `HEARTBEAT_OK` untouched, `workspace-loader.ts` untouched.

### ♻️ Reuse & Duplication (0)

No new util, component or abstraction — a string edit plus a test in an existing file, so there was no signal to search the codebase against.

### 🔴 Critical (0)

(none)

### 🟡 Medium (0)

(none)

### 🟢 Minor (2)

1. `src/orchestration/source-policy.ts:3` — the prohibition on writing `HEARTBEAT.md` is now only implied by "or other memory stores" instead of naming the file.
   → Low real risk: `AGENT_WRITABLE_FILES` (`workspace-loader.ts:433`) is MEMORY + IDENTITY + SKILL_LEARNING only and does not include `HEARTBEAT.md`, and an agent-role session is restricted to `mcp__gateway__*` tools (`process-turn.ts:146`), so it cannot write files at all. The prompt rule is a second layer here, not the only one.
2. `tests/unit/orchestration/persona-worker.test.ts:53` — the assertion pins the constant rather than the composed system prompt.
   → If someone later reintroduces `HEARTBEAT.md` through `SECRET_RULES` or `IDENTITY_EDIT_RULES` it would not be caught (both are clean today — 0 occurrences in the file). Asserting against the `sourceRules` string built at `bridge.ts:171` would cover more, at the cost of a heavier fixture. Acceptable trade-off.

### ✅ AC Coverage (#506)

- [x] `API_SOURCE_RULES` no longer contains `HEARTBEAT.md`
- [x] The memory-store prohibition still holds (3 assertions)
- [x] `HEARTBEAT_OK` untouched — not present in the diff
- [x] Regression test added
- [~] An API session whose agent says `HEARTBEAT_OK` completes normally — verified in the field (8/8 agents returned HTTP 200 after the workaround), but not covered by an automated end-to-end test, since that would require a live provider call.

### 🧪 Test evidence

`npx jest tests/unit/orchestration`, run on both branches:

- `main`: 10 failed suites, 5 failed tests, **505** passed
- this branch: 10 failed suites, 5 failed tests, **506** passed

Identical failure set on both (WhatsApp linked-account and MCP metadata discovery), pre-existing and caused by the `qrcode` / `pino` dependencies missing in this environment — the same root as all 10 `tsc --noEmit` errors, none of which are in `orchestration`. This branch adds exactly one passing test and breaks nothing.

PR #509, opened 2026-09-18
[Fix] Resume Unheard Browser Voice Replies After Navigation

0xMaxMa:
## Summary
Closes #508.

Preserve browser voice intent across transport detach and resume unheard approved speech using canonical response IDs, existing speech storage, the TTS queue and playback receipts.

- Runtime/store: persist per-session/principal voice intent and approved summaries while detached; query unheard completed responses with indexed, bounded catch-up.
- Voice API: authenticated text-mode preference endpoint, initially muted connections, periodic catch-up, and synchronous microphone lease reservation.
- VoiceSession: deduplicate live/catch-up speech, preserve completed playback on detach, and serialize rapid mute/unmute during finalization.
- Document the protocol and explicit-off versus navigation behavior.

## Verification
- `npm run build` passed.
- Voice/orchestration unit and integration suite: 110 suites, 841 tests passed.
- Latest voice API/session/resume suite: 32 tests passed; the final parameterized rapid-mute fixture also passed (26 session tests).
- Regression verified against the original VoiceSession: fully played response became `detached` instead of `played`; fixed version passes. Browser transport regressions were separately proven red in the companion client work.
- Real local HTTP/WebSocket test: simultaneous ticket requests yield 200/409; returning muted plays missed and later replies exactly once. Runtime/SQLite test verifies detached speech persistence and rejects another principal.
- EXPLAIN confirms session, response and delivery index searches.

## Manual checks / rollout
Deploy this gateway support before browser clients start using the preference endpoint and `voice.start.muted`.

1. Enable browser voice in session A, mute, navigate to B while an answer is pending, then return to A.
2. Confirm mute is retained and approved TTS plays for the unheard reply, including one completing after return.
3. Repeat navigation after playback completion; no duplicate speech should play.
4. Toggle mute/unmute during transcription; the committed turn should survive and the final microphone state should match the last action.
5. Explicitly turn voice off; reconnecting after enabling it again must not read text-mode backlog.

No physical microphone/provider/browser autoplay E2E or production deployment was performed. A partially played response interrupted by navigation restarts from its beginning; fully played or explicitly barged-in responses are excluded. Automatic catch-up can incur TTS usage. No full-report fallback is introduced.

PR #510, opened 2026-09-18
fix(voice): recover replay and confirm speech interruptions

0xMaxMa:
## Summary

Follow-up to #509, based on the reported develop session where approved speech was generated but interrupted before any acknowledged playback.

- Add authenticated, explicit POST replay generation from stored approved spoken text, including answers whose original playback was interrupted. Never fall back to arbitrary full response text.
- Keep the existing recording GET and `response_ids` contract; expose `replayable_response_ids` separately so buttons need not wait for retained audio.
- Bound replay synthesis to four concurrent jobs, 60 seconds and 16 MiB; share concurrent requests for the same scoped response and retain completed WAVs.
- Add acknowledged `playback.stop` for manual replay without cancelling agent work. Advance beyond unseen server playback epochs and echo the stop request ID.
- Add bounded `playback.pause` so STT classification can preserve buffered output beyond the normal client-progress timeout.
- Document the new controls and explicit replay generation API.
- Confirm interruption from final STT text even when the original response has already finished generating and the provider has no partial transcripts.

## Verification

- `npm run build` passed.
- 45 focused gateway tests passed across voice session, API, replay generation, browser resume and response streaming.
- Tests cover empty STT preserving playback, confirmed batch speech interrupting it, manual stop without task cancellation, session/owner isolation, missing-recording replay, original-byte reuse, concurrency and timeout bounds.
- A real WebSocket test reproduces the cleanup DELETE 404 after socket retirement; the companion frontend change handles that idempotently.

## Rollout and limitations

Companion GetPod PR: https://github.com/Crown-Labs/getpod/pull/2856. Deploy this gateway change before the new live replay control is used by the frontend. No live deployment has been performed. Real microphone/fan/echo testing remains necessary; automated tests use deterministic STT/TTS fixtures. Generating missing replay audio calls the configured TTS provider and uses its normal managed-credit checks.


0xMaxMa:
## Gateway review follow-up — 10c10c9

All findings from the requested review have been addressed on this PR.

### Findings resolved

- **Playback stop race:** explicit stop now advances past the current server epoch even when the browser has not received the latest `playback.start`. The server emits `playback.stopped` with a matching request ID and authoritative epoch. The companion client waits for it and discards intervening start/audio events.
- **STT classification pause:** `playback.pause` preserves buffered audio while the recognizer classifies a candidate. Backpressure and tail-drain waits account for a pause, bounded by utterance and provider finalization budgets. No fixed 1.2-second timer decides that speech is noise.
- **Documentation:** `website/api/voice.md` and README describe early replay, POST synthesis, billing, pause and stop acknowledgement controls.

### Evidence

- Build and 45 focused tests passed.
- Regression: a 15-second classification pause does not fail with `CLIENT_TOO_SLOW`, and output continues after resume.
- Regression: stop with an epoch equal to an unseen server start advances to a newer epoch and echoes the request ID.
- Executed the original cross-repository race reproduction using actual VoiceConnection and VoiceSession classes: both finish at epoch 2; stale live playback is rejected and explicit replay is not cancelled.

No additional blocker found in this follow-up. Actual fan/echo and microphone behavior still needs real-device testing with both PRs installed. Companion: Crown-Labs/getpod#2856.

PR #522, opened 2026-09-20
feat: mount optional host Codex runtime into app-agent containers

### LINKED ISSUE #521: [Enhancement] Mount optional host Codex runtime into app-agent containers
## Updated design decision: native Codex auth and readiness

Codex must use the user's own native provider/auth setup, just as Claude Code does. Do not implement Follow gateway/Claude auth, read Claude settings to supply Codex credentials, or introduce a central gateway credential store.

- Detect native executable/version, effective provider configuration and native auth availability under the actual gateway service user and HOME/CODEX_HOME. Native auth can be file-, keyring- or environment-backed; do not equate an absent auth.json with missing authentication.
- Distinguish local login state from verified provider access. Codex CLI 0.155.1 `login status` succeeds with a fake cached API key; real network/model readiness requires a separate bounded probe. Do not spend tokens on a probe every turn.
- In automatic worker selection, unavailable Codex falls back to Claude **before dispatch**, with an inspectable reason. Do not silently replay partially completed Codex work under Claude. Explicit safemode Codex requests should explain readiness failure and offer Claude; never change an existing native session's harness.
- Mount Codex runtime into app-agent containers only when the host executable and usable native auth are ready. Otherwise keep the Claude-only service operational. Reconcile eligibility after native login/logout/update at an idle boundary.
- Update the worker adapter as well as mounting: it currently creates an isolated CODEX_HOME and requires an independent configured API key. Preserve native provider/auth semantics while retaining scoped MCP/tool policy and workspace isolation. Do not expose the entire personal Codex home or use a permanently stale OAuth token copy; native refresh and concurrent host/container sessions require explicit tests.
- Gateway must not log users in, overwrite personal native configuration, or install Codex implicitly. Keep existing explicitly configured worker providers compatible or document an explicit migration.
- Add diagnostics and tests for no executable, no auth, native API-key/ChatGPT login, custom providers, credential rotation/expiry, keyring availability, unavailable network/quota, host/container execution and safemode resume.

GetPod installation/provisioning work is tracked in https://github.com/Crown-Labs/getpod/issues/2867; it prepares native CLI config/auth per tenant during VM/self-host provisioning instead of gateway borrowing Claude credentials at runtime.

These decisions supersede the earlier acceptance criterion that every missing-Codex task must fail: automatic routing now falls back before execution. Container auth isolation still applies; “no credential-directory mount” does not mean container workers may ignore native authentication.

## Summary
Make Codex an optional host-provided runtime for app-agent workers instead of installing a pinned Codex release into every generated agent image. Reuse a read-only host runtime when available, while keeping Claude-only apps independent of Codex.

## Investigation
Inspected commit `47f6bad601f8a8b9b397b18ce3ce1a00e90f0434`.

- `src/apps/agent-manager.ts:140` (`injectAgentService`) generates the agent image and Compose service. Lines 200–222 unconditionally download and checksum-verify Codex 0.154.0, including its bundled resources. Lines 248 onward already mount Claude, Node and npm packages read-only from the host.
- `src/session/codex-process.ts:225–234` chooses `config.bin ?? "codex"`. Host workers resolve that executable through the gateway process environment; app workers invoke it inside the app container using `docker exec`. Spawn errors become startup errors. There is no implicit fallback to Claude or host execution.
- `src/orchestration/worker-harness.ts` merges gateway/agent Codex settings and selects the harness. An explicit `workers.codex.bin` must remain supported.
- `src/cli/commands/doctor.ts:83` consumes dependency checks from `src/cli/dependencies.ts`; extend that shared mechanism for optional Codex diagnostics.
- `src/apps/agent-container-migration.ts:20` manages generated app-agent configuration migration and rollback. Runtime mount changes must integrate with this lifecycle without disturbing app data or active workers.
- `website/guide/worker-harnesses.md` currently documents image-installed Codex and image rebuild requirements.

Today the host and container can run different Codex versions, and even Claude-only apps download Codex during image builds. Host worker startup does not automatically install Codex.

## Proposed Approach
1. Remove unconditional Codex installation from generated `Dockerfile.agent` files. Retain existing Claude/Node behavior.
2. Introduce a shared runtime resolver honoring effective `workers.codex.bin` settings before gateway PATH discovery. Resolve symlinks and validate the executable. Support standalone and npm installations by locating the native runtime plus required bundled resources, not merely mounting an npm launcher or a dangling symlink. Define behavior for multiple agent overrides sharing one app container.
3. Generate read-only mounts to stable container runtime paths only when a compatible runtime is available. Do not mount the host Codex home, authentication, sessions, or unrelated host directories. Keep per-attempt container state and existing credential delivery isolated.
4. Keep app installation/startup working without Codex. When a task actually selects Codex and its runtime is missing or incompatible, fail promptly with a specific actionable error. Do not silently fall back to another harness or execute an app worker on the host.
5. Add optional dependency diagnostics to `doctor`, including resolved executable/version, service PATH differences and container compatibility. Do not silently install software; installation remains an explicit operator action.
6. Detect runtime changes after host upgrades or late installation. Reconcile generated mounts and recreate only the agent service at an appropriate idle boundary, or return explicit maintenance instructions. Never disrupt active workers, delete volumes, or recreate unrelated app/database services. Reuse migration backup/rollback protections where applicable.
7. Update worker-harness and doctor documentation with host prerequisites, override configuration, missing-runtime behavior and upgrade/recreation steps. Document Linux/architecture/runtime requirements and unsupported remote-Docker setups rather than assuming host paths are always visible to the daemon.

## Acceptance Criteria
- [ ] Generated agent images contain no unconditional Codex download/install step.
- [ ] A Claude-only app installs, starts and completes work with no Codex installed on the host.
- [ ] Codex host workers still honor explicit executable configuration and service PATH lookup.
- [ ] Codex app workers use the resolved host runtime through read-only mounts and remain inside the app container.
- [ ] Standalone and npm installs, symlinks, paths containing spaces, missing binaries, incompatible runtimes and explicit overrides have regression coverage.
- [ ] Missing Codex produces an actionable startup/task error and doctor finding; no silent harness or host fallback occurs.
- [ ] Host credentials/config/session directories are not exposed through runtime mounts; container auth/state isolation remains intact.
- [ ] Host upgrades and installing Codex after app creation are handled without stale mounts or unexpected interruption of active work.
- [ ] Existing generated app-agent deployments have a safe migration/reconciliation path with rollback and preserved data.
- [ ] Mock lifecycle tests and a container smoke test cover launch, MCP access within existing scope, cancellation and cleanup; Claude container behavior remains unchanged.
- [ ] Documentation reflects optional host provisioning and the actual supported installation/update workflow.

## Out of Scope
- Changing model routing defaults or silently falling back to Claude.
- Automatically installing Codex on the host without an explicit operator action.
- Expanding app-agent host permissions or mounting host personal Codex state.

0xMaxMa:
## Summary
Make Codex an optional host-provided worker runtime. Generated app-agent images no longer download a pinned Codex release; compatible host runtimes are mounted read-only while Claude-only installations remain supported without Codex.

Closes #521

## Changes
- Resolve effective worker executable overrides and service PATH, preserving workspace-relative paths and npm launchers for host workers.
- Discover native standalone/npm payloads and recognized bundled resources, reject incompatible architectures/layouts and escaping resource symlinks, and mount only runtime files at stable container paths.
- Require local Linux Docker with matching architecture. No remote Docker/Desktop assumption and no host Codex home, auth or session mounts.
- Verify selected mount fingerprint and mounted binary SHA-256 before Codex startup; fail explicitly on missing/stale/incompatible runtime without host or harness fallback.
- Add optional doctor diagnostics for resolved paths/version, effective per-agent overrides and container layout compatibility. No automatic Codex installation.
- Add local `app refresh-runtime NAME`: require a drained, stopped installer-owned agent, back up generated files, regenerate and recreate only `agent` using `--no-deps --no-build`, recheck container identity, and restore generated files on failure.
- Recover deleted old runtime mounts only with matching saved generated Compose definitions and fingerprint evidence. Existing automatic legacy migration retains its behavior.
- Update worker, maintenance and doctor documentation and the native container smoke fixture.

## Validation
- `npm run typecheck`: passed.
- `npm test -- --maxWorkers=2`: 441 suites / 7,020 tests passed on the final tree.
- Targeted tests: 11 suites / 254 tests passed; CLI generator synchronization: 2 tests passed.
- `cd mcp && bun test`: 45 tests passed.
- Native host and plain Debian container smoke tests passed: local fake Responses API, scoped MCP roundtrip, canonical summary, token accounting, native resume, mid-turn steering and cancellation. No provider billing.
- Original unconditional image-install regression was verified to fail on the old implementation. Refresh raw-config and workspace-relative-path regressions were also verified to fail with their pre-fix logic restored.
- The first full run caught generated CLI documentation drift. Updated the generator, regenerated the document, and reran the complete suite successfully without relaxing tests.

## Latest review fixes
- Docker context/info/inspect preflights now use asynchronous subprocesses with timeouts. Service generation and every installer/maintenance caller await completion, so slow Docker responses do not block gateway HTTP, voice or timers.
- Persist the immutable Docker container ID with native Codex session metadata. Same-container transcripts resume normally; recreation or a missing legacy transcript starts a fresh native conversation with the current task assignment and emits `native_session_reset`. Gateway task history is retained; native context from a deleted container cannot be recovered.
- Native container smoke now covers ordinary resume followed by recreation recovery, plus MCP, steering and cancellation.
- New regressions fail with pre-fix timer-blocking and session-recovery logic restored. Focused fix suites: 119 tests passed; typecheck and the complete 7,020-test suite passed. Independent review of the fixes found no additional confirmed findings.

## Review
Self-review corrected Docker Desktop detection and documented the stale-host-mount maintenance boundary. Independent review found that raw persisted app configs lacked the derived orchestration switch and that relative executable overrides changed their working directory; both were fixed with regression coverage. Follow-up independent review found no additional confirmed defects.

## Operational boundaries
- No running gateway or installed app was restarted, deployed or reconfigured during this work. Real Docker testing used only a temporary fixture container.
- Runtime changes return maintenance instructions; task startup never recreates containers. All agents sharing an app container must resolve to the same runtime.
- Refresh requires an existing generated agent image and recreates only the stopped agent; it does not build/install Codex, stop workers, delete volumes, or restart app/database services.
- Docker has no atomic stopped-state/Compose-recreate transaction. Operators must drain work and keep external launchers from starting the agent during maintenance.
- If refresh fails after recreation, files are restored and backups retained, but the old container is not automatically recreated. Inspect the error before explicit recovery.
- Removing host runtime files can also block Claude admission while unresolved mounts remain configured. Explicit stopped-agent refresh removes those mounts safely.
- Container compatibility targets official self-contained Linux x64/arm64 distributions. Arbitrary custom shared-library dependencies are not packaged automatically. Doctor checks layout; it does not certify every container image or service environment.


0xMaxMa:
### Review follow-up

Self-review and an independent code review are complete for `d602d64`.

Confirmed findings were fixed:
- Reject Linux Docker Desktop rather than treating every Unix endpoint as the local host namespace.
- Allow explicit stopped-agent refresh with normal raw persisted configuration, which lacks a derived orchestration switch.
- Preserve workspace-relative executable override semantics for host workers, mount generation and doctor diagnostics.

Follow-up review found no additional confirmed defects. Runtime removal and external-launcher maintenance boundaries are documented; refresh refuses active containers and never stops workers automatically.

Validation: 441 suites / 7,012 tests passed, typecheck passed, MCP 45 tests passed, and native host/container smoke tests passed. Regression tests were verified to fail with old image-generation, refresh-guard and relative-path logic. Generated CLI documentation drift found by the first full run was fixed at the generator and the complete suite rerun successfully.

0xMaxMa:
### Review fixes and re-review

Pushed `61dae04` addressing both findings:

1. Recreated app containers no longer resume vanished native transcript paths. Saved mappings include Docker container identity; same-container resume remains intact, while recreation or missing legacy transcripts starts a fresh native thread and emits a reset event. The current task assignment remains available; deleted native conversation history is not claimed to survive.
2. Docker context/info/inspect run asynchronously with timeouts, and all service-generation/install/refresh call sites await the result.

Independent re-review found no additional confirmed findings. Regression proof: restored old recovery logic fails two reset tests; restored synchronous preflight fails the timer-progress assertion. Fixed focused suites pass 119 tests. Final full suite: 441 suites / 7,020 tests passed; typecheck passed. Real disposable-container smoke passed same-container resume, recreation recovery, scoped MCP, steering and cancellation.

No running gateway or installed app was modified or restarted during this work.

